PR #4, opened 2021-04-08
[WIP] Add -p/--separate option to detach swayimg window from parent terminal

user18130814200115-2:
Feel free to make or suggest changes


user18130814200115-2:
I am still working on disabling all the calculations when -p is set to save some resources, but this is a proof of concept, let me know if you are interested in merging upstream.

artemsen (commented):
Actually, you don't need the `--separate` flag at all.
If you need a separate window you may disable Sway integration by setting `SWAYSOCK` environment variable:
```
SWAYSOCK="" swayimg myfile.jpg
```

user18130814200115-2:
Thank you so much for the quick and detailed response.
I have never actually written c code so your corrections are very helpful.

I see now that the --separate flag is not necessary and will close the pull request.

Thank you for the good work on this application.

PR #13, opened 2021-07-25
Make geometry CLI syntax match slurp

Seirdy:
The output of [`slurp`](https://github.com/emersion/slurp) describes
window geometry in the format `X,Y WIDTHxHEIGHT`; it's used by programs
like `grim` and `wf-recorder` already.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -17,7 +17,7 @@ static void print_help(void)
 {
     puts("Usage: " APP_NAME " [OPTION...] FILE...");
     puts("  -f, --fullscreen         Full screen mode");
-    puts("  -g, --geometry=X,Y,W,H   Set window geometry");
+    puts("  -g, --geometry=X,Y WxH   Set window geometry");
     puts("  -s, --scale=PERCENT      Set initial image scale");
     puts("  -i, --info               Show image properties");
     puts("  -v, --version            Print version info and exit");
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -17,7 +17,7 @@ static void print_help(void)
 {
     puts("Usage: " APP_NAME " [OPTION...] FILE...");
     puts("  -f, --fullscreen         Full screen mode");
-    puts("  -g, --geometry=X,Y,W,H   Set window geometry");
+    puts("  -g, --geometry=X,Y WxH   Set window geometry");
```
artemsen, line 20: It is not a common way to specify program arguments.
The "WxH" in this case is a next argument, it doesn't relate to geometry.
To fix this you have to escape a space or use quotes: `--geometry=X,Y\ WxH` or `--geometry="X,Y WxH"`.
Looks ugly for me =)

What about modifying `parse_rect()` to support both formats, or even better, to support any delimiters: commas, spaces, 'x', etc?

src/main.c
```diff
@@ -51,30 +51,47 @@ static void print_version(void)
...
  * @return false if rect is invalid
  */
 bool parse_rect(const char* arg, struct rect* rect)
 {
-    int* nums[] = { &rect->x, &rect->y, &rect->width, &rect->height };
-    size_t i;
     const char* ptr = arg;
-    for (i = 0; i < sizeof(nums) / sizeof(nums[0]); ++i) {
-        *nums[i] = atoi(ptr);
-        ptr = strchr(ptr, ',');
```
artemsen, line 66: We can handle here any delimiters (any not numeric values).

artemsen:
I expanded your idea slightly in the 3892fc6c12a97119d122b6ab1540a830ef2864ef.
Any delimiters are now allowed.

PR #20, opened 2022-02-13
Add support for system-level config (XDG_CONFIG_DIRS)

jirutka:


artemsen:
Hi Jakub,
Thank you for the patch!
Could you add `Signed-off-by` to your commits?

src/config.c
```diff
@@ -60,8 +60,8 @@ static FILE* open_file(void)
     size_t len;
     const char* postfix = "/swayimg/config";
     const size_t postfix_len = strlen(postfix);
+    char* config_dir = getenv("XDG_CONFIG_HOME");
```
artemsen, line 63: You don't need to remove `const`.

jirutka: Ah, right, `const` in C always confuses me.

jirutka:
I fixed the `const` and added `Signed-off-by`.

PR #23, opened 2022-02-14
Fix -Wimplicit-function-declaration on FreeBSD

jbeich:
According to [POSIX.1-2008](https://pubs.opengroup.org/onlinepubs/9699919799.2016edition/functions/gettimeofday.html) (compare with [POSIX.1-2001](https://pubs.opengroup.org/onlinepubs/009604599/functions/gettimeofday.html)) `gettimeofday` is `OB XSI` ("Obsolescent" + "X/Open System Interfaces"). XSI etxensions are provided by `_XOPEN_SOURCE`, a superset of `_POSIX_C_SOURCE`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -14,7 +14,7 @@ project(
 add_project_arguments(
   [
     '-Wno-unused-parameter',
-    '-D_POSIX_C_SOURCE=200809'
+    '-D_XOPEN_SOURCE=700'
   ],
   language: 'c',
 )


artemsen:
The `gettimeofday()` function is deprecated and I planned to replace it with a `clock_gettime()` call.
I don't mind this patch, but could you add `Signed-off-by` to your commit?

artemsen:
The `gettimeofday()` function replaced with `clock_gettime()`.

PR #24, opened 2022-02-14
Fix typo in libjxl detection

jbeich:
libjxl (unlike brunsli) always supported pkg-config: either `libjpegxl.pc` (before v0.1) or `libjxl.pc` (since v0.1). Otherwise, `-lfoo` expands to `libfoo.so` (or `libfoo.a`) but `-l:foo` expands to `foo`.
```
$ pkg info -x libjxl
libjxl-0.6.1_2

$ meson setup -Djxl=enabled _build
[...]
meson.build:35:0: ERROR: C shared or static library 'libjxl' not found
```


artemsen:
Looks like there is a problem with dependency declaration in `libjxl.pc`, it has `libhwy` as "requires", but `hwy` is needed only to build libjxl itself.
My system (Arch with libjxl 0.6.1-3) doesn't have `libhwy.pc` and pkconfig fails with:
```
$ pkg-config --cflags libjxl
Package libhwy was not found in the pkg-config search path.
```
As a result, meson cannot find `libjxl`:
```
The Meson build system
Version: 0.61.1
...
Run-time dependency libjxl found: NO (tried pkgconfig and cmake)
...
```
let's postpone this patch until Arch has a correct `libjxl.pc`.

jbeich:
`libhwy.pc` is provided by `highway`. Maybe Arch forgot to track dependency in `libjxl` package. On FreeBSD I've [patched](https://github.com/freebsd/freebsd-ports/blob/main/graphics/libjxl/files/patch-private-libhwy) `libjxl.pc` to avoid extra dependency.

Does `liblibjxl.so` (double `lib` prefix) actually exist on Arch? `jxl` is auto, so even if Arch package is broken it should be auto-disabled.

jbeich:
To fix Arch package one could move `highway` from `makedepends` to `depends`.

jbeich:
> `hwy` is needed only to build libjxl itself

If you're linking statically then linking against libhwy is also required. Generally, all dependencies of the library are transitively transfered to the consumer. pkg-config already thought of that, so there're `.private` extensions for keywords to help when `--static` is passed.

artemsen:
> So the real question would probably be "why, on some systems, is libjxl's dependency, libhwy, not installed as a dependency?"

Ok, you are right.
But I don't want to declare libhwy as a dependency of Swayimg, so...
Let's postpone this patch until Arch has ~~a correct libjxl.pc~~ libhwy installed as dependency of libjxl =)

artemsen:
Bug report in Arch: https://bugs.archlinux.org/task/73931

PR #29, opened 2022-06-15
Add option to specify a output display

artizirk:
This allows showing the swayimg window on specific display

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -12,6 +12,7 @@ typedef struct {
     uint32_t background; ///< Background mode/color
     bool fullscreen;     ///< Full screen mode
     bool show_info;      ///< Show image info
+    char* output_name;   ///< Output to show application window on
 } config_t;
 
 extern config_t config;
--- a/src/main.c
+++ b/src/main.c
@@ -19,6 +19,7 @@ static void print_help(void)
     // clang-format off
     puts("Usage: " APP_NAME " [OPTION...] FILE...");
     puts("  -f, --fullscreen         Full screen mode");
+    puts("  -o, --output=OUTPUT      Show the window on specific output");
     puts("  -s, --scale=TYPE         Set initial image scale: default, fit, or real");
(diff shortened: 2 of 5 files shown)


artemsen:
Hi Arti,
Thank you for the patch!
Could you, please, describe the scenario for using this feature?
As I can see, in such cases we don't need to add rules for setting the floating mode and absolute position of the window (but they are still added), all we need is a `move output`. But I am not sure we even need to create the output rule for each program instance, may be it would be better to create it only once in the sway config, something like `for_window [class="swaying*"] move output ...` and completely disable window management in swayimg. What do you think?

P.S. I am on vacation and far away from a computer this week, so my answer may be late, sorry ;)

artizirk:
> Could you, please, describe the scenario for using this feature?

So my usecase might be a bit weird. I want to be able to screenshot currently opened context menus with a screen crop (ie not the whole screen). Normal way to do this is to do a delayed screen capture because the context menu will close after it looses focus. But if I first do a full screen capture and then display that as an overlay, then I can use slurp to select the region and do the final screenshot of the selected region. swayimg seems to work really well for showing that temporary image.

    grim -o eDP-1 - | swayimg -f - &
    slurp | grim -g - && pkill swayimg

Only problem is that I have multiple displays and sometimes swayimg opens on a wrong screen. This patch fixes that. 

Other option would be to allow setting a custom app_id or class from cli arguments so that I could write rules to position multible swayimg windows on different displays.

artemsen:
@artizirk, I added support for custom app_id to the master branch:
`swaimg --app_id my_id_123`
If you decide to create own rule for swayimg in the Sway config, you may need to disable creating any additional and unneeded rules by swayimg:
`swaimg --app_id my_id_123 --no-rules`

Will this help?

artizirk:
Thank you! That will work for me. I will close this pull-request now.

PR #30, opened 2022-07-17
Improve filename completion

tsiflimagas:
Currently, when filename completion is being performed, dirnames don't have a slash appended to them and even worse, when completing dirnames, a space is appended instead of completing the contents of the directory. Using the `filenames` option, fixes this.

### DIFF
--- a/extra/bash.completion
+++ b/extra/bash.completion
@@ -21,6 +21,6 @@ _swayimg()
         COMPREPLY=($(compgen -f -- "${cur}"))
     fi
 } &&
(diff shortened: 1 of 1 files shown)


artemsen:
Hi Kostas,
Thanks you for the patch!
Could you add "Signed-off-by" to the commit message?

tsiflimagas:
Hi Artem, just force pushed the commit with the sign-off. Thanks for the quick response!

artemsen:
Thanks!

PR #55, opened 2022-10-18
Set the minimum required meson version to 0.59

b1rger:
The disable_auto_if method was introduced in 0.59
https://mesonbuild.com/Reference-manual_returned_feature.html#featuredisable_auto_if


artemsen:
Thank you!

PR #59, opened 2022-12-30
bash-completion: Use _filedir for space completion

ainola:
compgen -f doesn't play nicely with spaces, and bash-complete has since introduced _filedir which handles all of the gnarly work for file/dir handling.


artemsen:
Thank you!

PR #72, opened 2023-04-15
Add PNM support

abewieland:


src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+// Read an integer, ignoring leading whitespace
+static size_t pnm_readuint(struct pnm_iter* it)
+{
+    uint8_t c;
+    do {
+        c = pnm_readc(it);
+    } while (c == ' ' || c == '\t' || c == '\n' || c == '\r');
+    if (c < '0' || c > '9') longjmp(it->jmp, ERR_NONDIG);
+    size_t val = 0;
+    do {
```
artemsen, line 51: `strtoul`?

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+    }
+    plain = data[1] < '4';
+    type = (data[1] - '1') % 3;
+    it.data = data;
+    it.pos = 2;
+    it.size = size;
+
+    switch (setjmp(it.jmp)) {
+        case 0:
+            break;
```
artemsen, line 83: Continue even if we back?

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+                        g = div_near(g * 255, maxval);
+                        b = div_near(b * 255, maxval);
+                    }
+                    dst[x] |= ARGB_SET_R(r) | ARGB_SET_G(g) | ARGB_SET_B(b);
+                }
+            }
+        }
+    }
+
+    if (!type) {
```
artemsen, line 213: Please, merge these calls.

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+                    }
+                    dst[x] |= ARGB_SET_R(r) | ARGB_SET_G(g) | ARGB_SET_B(b);
+                }
+            }
+        }
+    } else {
+        // PGM and PPM (types 1 and 2, respectively) use 1-2 bytes for each
+        // channel depending on the maxval, with 1 channel for PGM and 3 for
+        // PPM; PBM (type 0) pads each row to the nearest whole byte
+        size_t bpc = maxval < 256 ? 1 : 2;
```
artemsen, line 163: Too many magic values. Please define constants.

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+        default:
+            image_print_error(ctx, "internal error");
+            return ldr_fmterror;
+    }
+
+    width = pnm_readuint(&it);
+    height = pnm_readuint(&it);
+    if (type) {
+        maxval = pnm_readuint(&it);
+        if (!maxval || maxval > 65535) {
```
artemsen, line 105: `UINT16_MAX` ?

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+// PNM loader implementation
+enum loader_status decode_pnm(struct image* ctx, const uint8_t* data,
+                              size_t size)
+{
+    bool plain;
+    int type;
+    struct pnm_iter it;
+    size_t width, height, maxval;
+
+    if (size < 2 || data[0] != 'P' || data[1] < '1' || data[1] > '6') {
```
artemsen, line 72: It would be better to define such constants as `1` and `6`.

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+        return ldr_unsupported;
+    }
+    plain = data[1] < '4';
+    type = (data[1] - '1') % 3;
```
artemsen, line 76: Does the PNM specification really uses `%` to specify the type? =)

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
+// SPDX-License-Identifier: MIT
```
artemsen, line 1: Please format the code according to .clang-format.

artemsen (changes requested):

src/formats/pnm.c
abewieland: Sorry, totally missed it! Will do once I make the larger structural changes above.

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+// PNM formats decoder
+// Copyright (C) 2023 Abe Wieland <abe.wieland@gmail.com>
+
+#include "loader.h"
+
+#include <setjmp.h>
+
+// Divide a by b, rounding to the nearest integer; evaluates b twice
+#define div_near(a, b) (((a) + (b) / 2) / (b));
```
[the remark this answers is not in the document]
abewieland: You're totally right; must have been going a little too much on autopilot there.

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+    return it->data[it->pos++];
+}
+
+// Read an integer, ignoring leading whitespace
+static size_t pnm_readuint(struct pnm_iter* it)
+{
+    uint8_t c;
+    do {
+        c = pnm_readc(it);
+    } while (c == ' ' || c == '\t' || c == '\n' || c == '\r');
```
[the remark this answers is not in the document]
abewieland: Sure; I originally had a version with `isspace()` and changed later on the basis that `isspace()` may change with different locales (and `libnetpbm` also had an explicit check), but to be fair the probability of a non C or POSIX locale is almost zero.

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+    uint8_t c = pnm_readc(&it);
+    if (c != ' ' && c != '\t' && c != '\n' && c != '\r') {
+        image_print_error(ctx, "single whitespace character expected");
+        return ldr_fmterror;
+    }
+    if (!image_create_frame(ctx, width, height)) {
+        return ldr_fmterror;
+    }
+
+    if (plain) {
```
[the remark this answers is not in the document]
abewieland: Will do! There's also a question of how much splitting is best; one could write a function for both plain and raw formats, or 6 functions for plain and raw versions of each of PBM, PGM, and PPM; splitting more involves more code duplication but may also be more efficient since there will be fewer branches in loops.

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+    size_t val = 0;
+    do {
+        uint8_t d = c - '0';
+        if (val > SIZE_MAX / 10) longjmp(it->jmp, ERR_RANGE);
+        val *= 10;
+        if (val > SIZE_MAX - d) longjmp(it->jmp, ERR_RANGE);
+        val += d;
+        c = pnm_readc(it);
+    } while (c >= '0' && c <= '9');
+    --it->pos;
```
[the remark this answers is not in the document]
abewieland: The `--it->pos` is there to "unput" the most recent character (like `ungetc`) so that future functions will work with the first non-digit character. I guess I could have put it in a function, but that function feels a little trivial in that case.

src/formats/pnm.c
abewieland: Unfortunately `strtoul` is unusable for two reasons:
First, comments in PNM files may appear anywhere (even in other tokens). This fact needlessly complicates this parser, but we need to deal with it in some way to handle all possible PNM files.
Second, there isn't a guarantee the buffer is null-terminated, and `strtol` and friends don't allow a max length to be specified.

src/formats/pnm.c
```diff
@@ -0,0 +1,222 @@
...
+
+#include <setjmp.h>
+
+// Divide a by b, rounding to the nearest integer; evaluates b twice
+#define div_near(a, b) (((a) + (b) / 2) / (b));
+// Divide a by b, rounding up; evalutes b twice
+#define div_ceil(a, b) (((a) + (b) - 1) / (b));
+
+// A file-like abstraction with error handling via setjmp/longjmp which
+// significantly helps code readability
```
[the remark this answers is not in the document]
abewieland: I agree that in general `setjmp` is terrible for code readability; I only used it here after discovering the alternatives were not great. My main reason for doing so was that we have a fairly small number of errors in parsing, all of which should halt the entire operation, but those errors are detected in different functions. Thus, it actually seemed pretty clean to have errors all land in a single place (which would then print the error and return `ldr_fmterror`) rather than propagating return values.

It also seems reasonable to rewrite `pnm_readc` and `pnm_readuint` to have an error code return and write to pointer arguments, but this would simply make the functions which call them end up having either a lot of repeated code along the lines of
```
size_t v;
int r = pnm_readuint(&it, &v);
if (r) {
    image_print_error(...)
    return ldr_fmterror;
}
```
Honestly I would probably write these checks as local `goto`s anyway (to avoid writing the error condition and return code every time) so a global goto also seemed reasonable, but I can understand your aversion.

src/formats/pnm.c
abewieland: I'm not entirely sure what you mean here. `setjmp` returns 0 the very first time it's called (to set up the return location); every return from a `longjmp` call returns non-zero. It might have been more idiomatic to write
```
int r = setjmp(it.jmp);
if (r) {
    switch (r) { ... }
}
```
Since most of the time one sees code like `if(setjmp(...)) { /* handle error */ }`; I can definitely change it to that if desired.

src/formats/pnm.c
abewieland: Well, `P1` and `P4` are both PBM files, `P2` and `P5` are both PGM, and `P3` and `P6` are both PPM files. The lower number for each is the ASCII/plain representation, while the higher number is the raw/binary representation. If you prefer, I could definitely write something like
```
#define PBM_PLAIN '1'
...
#define PPM_RAW   '6'
...
#define TYPE_PBM   0
#define TYPE_PGM   1
#define TYPE_PPM   2
...
switch (data[1]) {
    case PBM_PLAIN:
        plain = true;
        type = TYPE_PBM;
        break;
    ....
    case PPM_RAW:
        plain = false;
        type = TYPE_PPM;
        break;
}
```

src/formats/pnm.c
abewieland: I'm really sorry, I guess I hadn't done enough research. The standard states that comments may appear anywhere, including inside tokens; they even mention how weird this is:
> Before the whitespace character that delimits the raster, any characters from a "#" through the next carriage return or newline character, is a comment and is ignored. Note that this is rather unconventional, because a comment can actually be in the middle of what you might consider a token.

However, every parser that I've since looked at in more depth (`libnetpbm`, `gdk-pixbuf`, `libfreeimage`, and `imagemagick`) disallow comments within tokens (they simply start new tokens, as would have been the sane thing to have in the standard). This simplifies matters, though it still leaves `strtol` in an unusual spot (due to the lack of length checking); I think I'll probably still do it manually (like `libnetpbm`, `libfreeimage`, and `imagemagick`), though I could copy to an external buffer (null-terminated) and call `strtol` on that (like `gdk-pixbuf`).

src/formats/pnm.c
```diff
@@ -0,0 +1,287 @@
...
+            image_print_error(ctx, pnm_strerror(maxval));
+            return ldr_fmterror;
+        }
+        if (!maxval || maxval > UINT16_MAX) {
+            image_print_error(ctx, "invalid maxval");
+            return ldr_fmterror;
+        }
+    }
+    if (!plain) {
+        if (*it->pos != ' ' && *it->pos != '\t' && *it->pos != '\n' &&
```
artemsen, line 263: Ok, I see you don't like `isspace`, I don't mind ))
Let's reduce this line in this way:
```
const char c = *it->pos;
if (c != ' ' ...
```
And the same story in `pnm_readint`.

src/formats/pnm.c
```diff
@@ -0,0 +1,287 @@
...
+            dst[x] = pix;
+        }
+    }
+    return 0;
+}
+
+enum loader_status decode_pnm(struct image* ctx, const uint8_t* data,
+                              size_t size)
+{
+    struct pnm_iter it[1];
```
artemsen, line 203: `struct pnm_iter it;` is enough.

src/formats/pnm.c
artemsen: Ok, let's leave it as is.

src/formats/pnm.c
```diff
@@ -0,0 +1,287 @@
...
+            break;
+        default:
+            return ldr_unsupported;
+    }
+    it->pos = data + 2;
+    it->end = data + size;
+
+    width = pnm_readint(it);
+    if (width < 0) {
+        image_print_error(ctx, pnm_strerror(width));
```
artemsen, line 243:
```
../src/formats/pnm.c: In function ‘decode_pnm’:
../src/formats/pnm.c:243:9: warning: format not a string literal and no format arguments [-Wformat-security]
  243 |         image_print_error(ctx, pnm_strerror(width));
      |         ^~~~~~~~~~~~~~~~~
```

Use instead:
```
image_print_error(ctx, "%s", pnm_strerror(width));
```

src/formats/pnm.c
```diff
@@ -0,0 +1,287 @@
...
+    if (it->pos == it->end)
+        return PNM_EEOF;
+    if (*it->pos < '0' || *it->pos > '9')
+        return PNM_EFMT;
+    int val = 0;
+    // XXX strtol is still on the table, but to do so we'd need to make a first
+    // pass to copy to a buffer, so it seems easier to stick with the majority
+    // and do this ourselves as well
+    do {
+        uint8_t d = *it->pos - '0';
```
artemsen, line 76: Can be `const`

src/formats/pnm.c
```diff
@@ -0,0 +1,287 @@
...
+            argb_t pix = ARGB_SET_A(0xff);
+            if (type == PNM_PBM) {
+                int bit = pnm_readint(it);
+                if (bit < 0)
+                    return bit;
+                if (bit > 1)
+                    return PNM_EOVF;
+                pix |= bit - 1;
+            } else if (type == PNM_PGM) {
+                int v = pnm_readint(it);
```
artemsen, line 106: Can be `const`.
Please, add the `const` qualifier to read-only variables in other places.

src/formats/pnm.c
```diff
@@ -0,0 +1,287 @@
...
+        if (val > INT_MAX - d)
+            return PNM_ERNG;
+        val += d;
+        ++it->pos;
+    } while (it->pos != it->end && *it->pos >= '0' && *it->pos <= '9');
+    return val;
+}
+
+// Decode a plain/ASCII PNM file; returns 0 on success, an error code otherwise
+static int decode_plain(struct image_frame* f, struct pnm_iter* it, int type,
```
artemsen, line 91: It would be better to use `enum` instead of `int` for the type. It will give us a type check in compile time.

src/formats/pnm.c
```diff
@@ -0,0 +1,287 @@
...
+        argb_t* dst = f->data + y * f->width;
+        const uint8_t* src = it->pos + y * rowsz;
+        for (size_t x = 0; x < f->width; ++x) {
+            argb_t pix = ARGB_SET_A(0xff);
+            if (type == PNM_PBM) {
+                int bit = (src[x / 8] >> (7 - x % 8)) & 1;
+                pix |= bit - 1;
+            } else if (type == PNM_PGM) {
+                int v;
+                if (bpc == 1)
```
artemsen, line 165: Please, add `{}` for `if` when you have more then 2 lines in a block (include else). I prefer squares even for one-lined if's.

artemsen (changes requested):
Please include a brief description of the PNM format in the commit message, we need at least a link to the spec.

src/formats/pnm.c
abewieland: Yeah, I use `setjmp` rarely enough myself that I have to check the manpage every time I see it in code. With the removal of comments in tokens, though, it wasn't too bad to get rid of it altogether.

src/formats/pnm.c
abewieland: Sure! I had done so to allocate `it` but ensure it's a pointer, allowing me to avoid some of the address-of operators when passing to functions, but I'm pretty ambivalent on this point.

src/formats/pnm.c
abewieland: Sorry, I wasn't getting this warning, so didn't know I should look at it. I've now compiled with `gcc` and `clang` for both build type `debug` and `release` with no warnings, so hopefully we're in the clear.

PR #83, opened 2023-06-01
Add MIME types to desktop entry

tasmo:
- make the integration into graphical file managers easier
- additional generic name
- set not to display in menus

Reference: https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html


artemsen:
Thank you!

PR #86, opened 2023-07-01
Support for larger BMP DIB headers

voidbert:
### What was added

Added support for BMP images with DIB headers larger than the standard `BITMAPINFOHEADER`. An example is `BITMAPINFOV5HEADER`, that GIMP outputs.

### How it was added

Consider the figure below. Before, the mask was read from the same place as the color data (purple), immediately after the DIB header. However, mask information is contained **in** the DIB header, in versions that support it. 

image

A fix was added to detect if the DIB header contains mask information (according to its size, see [this](https://formats.kaitai.io/bmp/)), and load it accordingly.

### DIFF
--- a/src/formats/bmp.c
+++ b/src/formats/bmp.c
@@ -26,6 +26,14 @@
 #define MASK555_RED   0x001f
 #define MASK555_GREEN 0x03e0
 #define MASK555_BLUE  0x7c00
+#define MASK555_ALPHA 0x0000
+
+// Sizes of DIB Headers
+#define BITMAPINFOHEADER_SIZE 0x28
+#define BITMAPINFOV2HEADER_SIZE 0x34
+#define BITMAPINFOV3HEADER_SIZE 0x38
+#define BITMAPINFOV4HEADER_SIZE 0x6C
+#define BITMAPINFOV5HEADER_SIZE 0x7C
 
 #define BITS_PER_BYTE 8
 
@@ -57,6 +65,7 @@ struct __attribute__((__packed__)) bmp_mask {
     uint32_t red;
     uint32_t green;
     uint32_t blue;
+    uint32_t alpha;
 };
 
 // Color palette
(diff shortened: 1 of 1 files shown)


src/formats/bmp.c
```diff
@@ -355,7 +371,8 @@ enum loader_status decode_bmp(struct image* ctx, const uint8_t* data,
     const void* color_data;
     size_t color_data_sz;
     struct bmp_palette palette;
-    const struct bmp_mask* mask;
+    const uint32_t* mask_location;
```
artemsen, line 374: It can be moved inside the `if` block to reduce the variable scope.

src/formats/bmp.c
```diff
@@ -383,11 +400,27 @@ enum loader_status decode_bmp(struct image* ctx, const uint8_t* data,
     color_data_sz = hdr->offset - sizeof(struct bmp_file) - bmp->dib_size;
     palette.table = color_data;
     palette.size = color_data_sz / sizeof(uint32_t);
-    mask = (color_data_sz <= sizeof(*mask) ? color_data : NULL);
+
+    // create mask
+    if (bmp->dib_size > BITMAPINFOHEADER_SIZE) {
+        mask_location = (const uint32_t*)(bmp + 1);
+
+	  mask.red = mask_location[0];
```
artemsen, line 408: Please, reformat the code with project's [.clang-format](https://github.com/artemsen/swayimg/blob/master/.clang-format), this will remove tab/space mix.

src/formats/bmp.c
```diff
@@ -383,11 +400,27 @@ enum loader_status decode_bmp(struct image* ctx, const uint8_t* data,
...
+	  mask.green = mask_location[1];
+	  mask.blue = mask_location[2];
+
+        if (bmp->dib_size > BITMAPINFOV2HEADER_SIZE) {
+            mask.alpha = mask_location[3];
+        } else {
+            mask.alpha = 0;
+        }
+    } else {
```
artemsen, line 417: I would swap if/else blocks, this will make them more readable: short one-lined `if` and then long `else`.

artemsen (changes requested):
Thank you for the patch!
But I tried to test it and found that it breaks the existing functionality.
I used the Bitmap Test Suite from https://bmptestsuite.sourceforge.io, at least this image is not decoded:
[32bpp-101110-320x240.bmp.tar.gz](https://github.com/artemsen/swayimg/files/11928912/32bpp-101110-320x240.bmp.tar.gz)

src/formats/bmp.c
voidbert: I don't mind trying to fix those problems. I tested that test suite (and others, namely [this one](https://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html)) before, and I didn't realize that particular image was broken, my bad.

I must note that many other images are (and were already) broken, namely RLE images with delta codes. I am someone who loves reinventing the wheel and writing my own implementations for things that already have one. However, even if the BMP format seems simple, it certainly isn't as simple as people think. Have you considered using a library for perfect compatibility?

artemsen: > I am someone who loves reinventing the wheel and writing my own implementations
Yeah, me too =)

> even if the BMP format seems simple, it certainly isn't as simple as people think
Exactly, I've spent a lot of time writing the current implementation and it still doesn't cover all things.

> Have you considered using a library for perfect compatibility?
This should be a library that is available on most Linux distributions and has a C interface.
I didn't find anything suitable. That's why I have my own implementation.

Issue #102, opened 2023-12-27
More Zooming & Panning Shortcuts


SpaceshipOperations:
Hello, first of all thank you for this nice app!

I would like to suggest a few really nice-to-have QoL shortcuts.

## Zooming Shortcuts

- Fit width: Fits image width to window width (even if the height is gonna be larger).
- Fit height: Fits image height to window height (even if the width is gonna be larger).
- Toggle zoom: Toggles between `zoom_fit` and `zoom_real`. More specifically:
    + If current zoom level is "fit", it switches to "real".
    + If current zoom level is "real", it switches to "fit".
    + If current zoom level is neither "fit" nor "real", it switches to "fit".

## Panning Shortcuts

It would be great to have the ability to configure multiple panning shortcuts with a different step size. For example it would be ideal for me to configure it like this:

```ini
[keys]
# Small steps
Up = step_up 1%
Down = step_down 1%
Left = step_left 1%
Right = step_right 1%

# Large steps
Shift+Up = step_up 30%
Shift+Down = step_down 30%
Shift+Left = step_left 30%
Shift+Right = step_right 30%
```

Likewise it would be great to be able to do the same for zooming:

```ini
[keys]
# Small step
equal = zoom_in 10%
minus = zoom_out 10%
# Large steps (with shift)
plus = zoom_in 25%
underscore = zoom_in 25%
```

Honestly if these are implemented, I would uninstall Gwenview forever lol.

artemsen:
> Fit width: Fits image width to window width (even if the height is gonna be larger).
> Fit height: Fits image height to window height (even if the width is gonna be larger).

Looks like it is [zoom_fill](https://github.com/artemsen/swayimg/commit/d0a4a0ed46f01bd953536f3421f3baaa0ff27cf7) mode, `Z` key by default.

> Toggle zoom: Toggles between zoom_fit and zoom_real

For now, we have `zoom_optimal`, `zoom_fit`, `zoom_fill`, and `zoom_real` modes. You can bind them to any keys. Is it enough?

> It would be great to have the ability to configure multiple panning shortcuts with a different step size.
> Likewise it would be great to be able to do the same for zooming

I [did it](https://github.com/artemsen/swayimg/commit/4564487101c9b064a102caecfabe1e9b2f9777cb) 4 day ago =)
```
[keys]
Left = step_left 10
a = step_left 50
plus = zoom_in 10
b = zoom_in 20
```

But all these features are in the `master` branch and have not yet been released.
[...]

PR #104, opened 2023-12-29
Add fit-width and fit-height options.

SpaceshipOperations:
This PR implements the "fit width" and "fit height" features mentioned in issue #102, along with their corresponding configuration options, and a couple of suggested default key bindings.

The commit is signed as per the convention in the repository. I compiled and tested the new features and they work as expected.

If you would like me to make any adjustments (such as removing the default bindings or changing the enum variable names), I can gladly do so. Or you can simply make those adjustments after accepting the PR.

Thank you in advance!


artemsen:
Thank you!
I'm just now refactoring this part of the code to get rid of duplication of structures.
Therefore, please note that the configuration format will be changed in the release version.

SpaceshipOperations:
Thanks!

PR #111, opened 2024-01-14
Add support for 2-dimensional scrolling with touch pads.

SpaceshipOperations:
This PR adds support for 2-dimensional scrolling, which is useful for touch
pad users.

In order to facilitate, the keyboard symbols that scroll events map into were
changed into the keypad up/down/left/right keys (so now we have four symbols
rather than two), and default bindings were added for those symbols so that
two-dimensional touch pad scrolling works out-of-the-box.


src/ui.c
```diff
@@ -28,6 +28,16 @@
 #define BTN_LEFT 0x110 // from <linux/input-event-codes.h>
 #endif
 
+#ifndef WL_POINTER_AXIS_ENUM
```
artemsen, line 31: Do we really need this block?

src/ui.c
```diff
@@ -349,7 +359,11 @@ static void on_pointer_button(void* data, struct wl_pointer* wl_pointer,
 static void on_pointer_axis(void* data, struct wl_pointer* wl_pointer,
                             uint32_t time, uint32_t axis, wl_fixed_t value)
 {
-    xkb_keysym_t key = value > 0 ? XKB_KEY_SunPageDown : XKB_KEY_SunPageUp;
+    xkb_keysym_t key;
+    if (axis == WL_POINTER_AXIS_HORIZONTAL_SCROLL)
+        key = value > 0 ? XKB_KEY_KP_Right: XKB_KEY_KP_Left ;
```
artemsen, line 364: Not enough one space and one space is excessive. Use clang-format =)

src/ui.c
```diff
@@ -349,7 +359,11 @@ static void on_pointer_button(void* data, struct wl_pointer* wl_pointer,
...
+    else
+        key = value > 0 ? XKB_KEY_KP_Down : XKB_KEY_KP_Up;
```
artemsen, line 366: Using keys for touch/mouse is a dirty hack.
I think we need own codes for that, may be not in this PR, but in the future.

artemsen:
Please, remove the dot from the commit subject, put your first comment into commit message, and add `Signed-off-by` to last line.

artemsen (changes requested):

SpaceshipOperations:
>Please, remove the dot from the commit subject, put your first comment into commit message, and add Signed-off-by to last line.

Done.

>Not enough one space and one space is excessive.

Fixed.

>Do we really need this block?

I mean, personally I would've preferred to just `#include` the header, but I noticed you defined `BTN_LEFT` manually instead of including its header, so I maintained the same style. If you want I can remove the block and put an `#include` line instead.

>Using keys for touch/mouse is a dirty hack.
>I think we need own codes for that, may be not in this PR, but in the future.

I'm aware. I'm considering doing that myself, but I'd need to dig up some documentation/examples, so for the time being I went with the absolutely bare minimum modification lol.

artemsen:
Thanks!

PR #115, opened 2024-02-05
set project version to 2.1

onlyJakob:
set project version to 2.1 in meson_options.txt

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -67,5 +67,5 @@ option('desktop',
 # project version
 option('version',
        type : 'string',
-       value : '0.0.0',
+       value : '2.1',
        description : 'project version')


artemsen:
This parameter should be set during the build process. I don't want to change this line on every release =)

artemsen:
Also, this patch breaks the version info because of [this](https://github.com/artemsen/swayimg/blob/2b5cee5ee477e886a2d3189a7e6a2c018f84e084/meson.build#L26).

onlyJakob:
Okay, I didn't see that.

maybe this should be mentioned in the readme, at least in [openSUSE Tumleweed](https://build.opensuse.org/package/show/openSUSE:Factory/swayimg) the version is not set correctly during buildtime. I have submitted a patch there :)

artemsen:
> maybe this should be mentioned in the readme

Maybe. But this is a very specific info just for maintainers.

> in openSUSE Tumleweed the version is not set correctly during buildtime

But this is problem of openSUSE, not a `swayimg` one =)
In [AUR](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=swayimg-git#n7), for example, I set it manually.

PR #118, opened 2024-02-09
Improve alphabetical sorting performance

aaronayub:
This replaces the current algorithm for alphabetical sorting of paths with qsort from the C standard library. This improves sorting performance, allowing swayimg to start up quicker on large directories with the recursive option enabled.


artemsen:
Thank you!

PR #141, opened 2024-06-11
Fix mouse pan

diniamo:
Using integers for coordinates and mouse delta doesn't work because both dx and dy would and up being either -1 or 1, which only allows diagonal panning.

Keep in mind that I only changed the types in the relevant places, and while this compiles fine, I'm pretty sure it's incomplete, please point out anything else this needs.

### DIFF
--- a/src/ui.c
+++ b/src/ui.c
@@ -331,7 +331,7 @@ static void on_pointer_motion(void* data, struct wl_pointer* wl_pointer,
     if (ctx.mouse.active) {
         const int dx = x - ctx.mouse.x;
         const int dy = y - ctx.mouse.y;
-        if (dx && dy) {
+        if (dx || dy) {
             viewer_on_drag(dx, dy);
         }
     }


src/ui.c
```diff
@@ -325,13 +326,13 @@ static void on_pointer_motion(void* data, struct wl_pointer* wl_pointer,
...
+    const double x = wl_fixed_to_double(surface_x);
+    const double y = wl_fixed_to_double(surface_y);
 
     if (ctx.mouse.active) {
-        const int dx = x - ctx.mouse.x;
-        const int dy = y - ctx.mouse.y;
-        if (dx && dy) {
+        const double dx = x - ctx.mouse.x;
+        const double dy = y - ctx.mouse.y;
+        if (dx || dy) {
```
artemsen, line 335: Good catch! This is indeed a typo.
But we don't need floating points here. Swaying doesn't use OpenGl, Vulkan or any other 3D API, so everything related to the window is measured in pixels.
I checked both variants, there is no difference.
So, it is enough to fix only this line. And add Signed-off-by, please.

diniamo: Ah you're right, I should've realized that actually.

I've got a question actually: what's size_t and ssize_t? I couldn't find any good explanations.

artemsen: size_t is an unsigned number that has the size of a pointer, that is, the size of a general CPU register: 64 bits on x86_64.
ssize_t is its signed alternative.

PR #142, opened 2024-06-15
Add timeout option for image info block

nicolasbrailo:
Create a new timer in viewer to hide image info blocks on trigger; when set, image info will be displayed on screen until the timeout expires. When the timeout expires, the UI will redraw and hide info blocks.

The timeout is disabled by default (info block shown forever) but a user may override it to a fixed time (in seconds) or to a percentage of a frame time in slideshow mode (eg if slideshow_time=10 and image_info_timeout=50%, the info block will be hidden after 5 seconds).

## Test in slideshow mode

Before timeout:
image

After timeout:
image

## Also tested
* non slideshow mode: info block is hidden after timeout, but only when timeout is configured in seconds
* default behaviour: info block is never hidden
* timeout bigger than slideshow time: info block is never hidden

### DIFF
--- a/extra/swayimgrc
+++ b/extra/swayimgrc
@@ -75,6 +75,8 @@ shadow = #000000
 [info]
 # Mode on startup (off/brief/full)
 mode = full
+# Hide image information after a timeout (seconds e.g. `5`, or percent of slideshow_time e.g. `50%`)
+timeout = 50%
 # Display scheme for the "full" mode: position = content
 full.topleft = name,format,filesize,imagesize,exif
 full.topright = index
--- a/src/info.c
+++ b/src/info.c
@@ -107,6 +107,7 @@ struct info_block {
 /** Info data context. */
 struct info_context {
     enum info_mode mode;
+    int timeout;
     const char* file;
     struct info_line* exif_lines;
     size_t exif_num;
@@ -207,6 +208,24 @@ static enum config_status load_config(const char* key, const char* value)
     size_t scheme_sz = 0;
     ssize_t index;
 
+    if (strcmp(key, VIEWER_CFG_INFO_TIMEOUT) == 0) {
+        size_t val_sz = strlen(value);
+        bool timeout_is_rel = false;
+        if (value[val_sz - 1] == '%') {
+            val_sz = val_sz - 1;
+            timeout_is_rel = true;
+        }
+
+        ssize_t num;
+        const ssize_t maxVal = timeout_is_rel ? 100 : 86400;
+        if (str_to_num(value, val_sz, &num, 0) && num != 0 && num <= maxVal) {
+            ctx.timeout = num * (timeout_is_rel ? -1 : 1);
+            return cfgst_ok;
+        } else {
+            return cfgst_invalid_value;
+        }
+    }
+
     if (strcmp(key, "mode") == 0) {
         index = str_index(mode_names, value, 0);
         if (index < 0) {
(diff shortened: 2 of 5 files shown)


artemsen:
I like both the idea and implementation, thank you!

But I think it would be better to move the option `image_info_timeout` into the `[info]` section.
And please add "Signed-off-by" to the commit message.

nicolasbrailo:
> But I think it would be better to move the option image_info_timeout into the [info] section.

Sure. @artemsen is it possible to observe two different config sections via `config_add_loader`? That is, can I add a second `config_add_loader` for `CONFIG_SECTION` in viewer.c, without breaking anything, or is it better to expose the new config setting through a method in info.h?

artemsen:
> is it possible to observe two different config sections

Yes, it is possible, but I would suggest to extend the info interface, something like this:
```
/**
  * Get info display timeout. 
  * @return 0 if timeout disabled, positive number for absolute time in second, or negative number for slideshow relative percents.
  */
int info_timeout(void);
```

nicolasbrailo:
Sounds simple enough. Would you prefer a change on top of this one, or a
new PR with all changes squashed into a single commit?

artemsen:
I would prefer to have one squashed commit in this PR.
But currently I am on vacation, so I will return to review in July, sorry :)

extra/swayimgrc
```diff
@@ -75,6 +75,8 @@ shadow = #000000
 [info]
 # Mode on startup (off/brief/full)
 mode = full
+# Hide image information after a timeout (seconds e.g. `5`, or percent of slideshow_time e.g. `50%`)
+timeout = 50%
```
artemsen, line 79: This file contains default settings, so it should be `timeout = 0`

src/info.c
```diff
@@ -207,6 +208,24 @@ static enum config_status load_config(const char* key, const char* value)
     size_t scheme_sz = 0;
     ssize_t index;
 
+    if (strcmp(key, VIEWER_CFG_INFO_TIMEOUT) == 0) {
+        size_t val_sz = strlen(value);
+        bool timeout_is_rel = false;
+        if (value[val_sz - 1] == '%') {
```
artemsen, line 214: Segfault if `value` is empty (`val_sz == 0`)

src/info.h
```diff
@@ -6,6 +6,9 @@
 
 #include "font.h"
 
+// Config key for info block timeout
+#define VIEWER_CFG_INFO_TIMEOUT "timeout"
```
artemsen, line 10: It is info's object, so `INFO_CFG_TIMEOUT` and it would be better to move it to `info.c`.

src/viewer.c
```diff
@@ -340,12 +345,30 @@ static void reset_state(void)
...
     slideshow_ctl(ctx.slideshow_enable);
+
+    // Expire info block after timeout
+    const int timeout = info_timeout();
+    if (timeout != 0) {
+        struct itimerspec info_ts = { 0 };
+        if (!ctx.slideshow_enable && timeout < 0) {
+            // If timeout is relative to slideshow, but slideshow is not
+            // enabled, then disable info timeout too
+            info_ts.it_value.tv_sec = 0;
```
artemsen, line 363: It is already `0` (see line 359).

src/viewer.c
```diff
@@ -588,6 +623,13 @@ void viewer_init(void)
         ui_add_event(ctx.slideshow_fd, on_slideshow_timer);
     }
 
+    // setup info block timer
+    ctx.info_timeout_fd =
+        timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC | TFD_NONBLOCK);
```
artemsen, line 628: We don't have to create timer if it disabled in config.

artemsen (commented):
Please, add `Signed-off-by` to the commit message.

artemsen:
I'm refactoring the image list and the loader, so it's easier for me to fix it myself.
Thank you for PR!

PR #143, opened 2024-06-16
allow for translucent colors on window and image backgrounds

an-prata:
Allows for colors in the format `#RRGGBBAA` in addition to the already permitted `#RRGGBB` for window and transparent image backgrounds. This change makes the `none` config option exactly equivilent to `#00000000`, and both are acceptable - this PR does not (to my knowledge) make any breaking changes. This format is disallowed for other config options still.

Implemented via a slightly modified `config_to_color` function called `config_to_translucent_color`, which will make the numeric conversion from hex to a 32 bit integer as usual, then masks out the alpha and shifts some bits to match the format of `argb_t` if the input string is longer than 6 characters.

Since the alpha channel was used for a sentinal value indicating a grid background for images - the indication of this option is now a `bool` in `viewer.c`'s `viewer` struct.

Am happy to hear feedback in case any changes are sub-optimal or have a prefered alternative - thanks!

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -4,6 +4,7 @@
 
 #include "config.h"
 
+#include "src/pixmap.h"
 #include "str.h"
 
 #include <ctype.h>
@@ -297,6 +298,30 @@ bool config_to_color(const char* text, argb_t* color)
     }
 
     if (str_to_num(text, 0, &num, 16) && num >= 0 && num <= 0xffffffff) {
+        *color = num;
+        return true;
+    }
+
+    return false;
+}
+
+bool config_to_translucent_color(const char* text, argb_t* color)
+{
+    ssize_t num;
+
+    if (*text == '#') {
+        ++text;
+    }
+
+    if (str_to_num(text, 0, &num, 16) && num >= 0 && num <= 0xffffffff) {
+        if (num > 0x00ffffff) {
+            ssize_t alpha = 0x000000ff & num;  // extract alpha
+            num >>= 8;                         // shift r, g, and b into expected position
+            num |= ARGB_SET_A(alpha);          // recompose
+        } else {
+            num |= ARGB_SET_A(0xff);
+        }
+
         *color = num;
         return true;
     }
--- a/src/config.h
+++ b/src/config.h
@@ -65,9 +65,17 @@ void config_add_loader(const char* section, config_loader loader);
 bool config_to_bool(const char* text, bool* flag);
 
 /**
- * Convert text value to ARGB color.
+ * Convert text value to ARGB color, with a constant max value for alpha.
  * @param text text to convert
  * @param color output variable
  * @return false if text has invalid format
  */
 bool config_to_color(const char* text, argb_t* color);
+
+/**
+ * Convert text value to ARGB color which may be any alpha.
+ * @param text text to convert
+ * @param color output variable
+ * @return false if text has invalid format
+ */
+bool config_to_translucent_color(const char* text, argb_t* color);
--- a/src/viewer.c
+++ b/src/viewer.c
@@ -21,9 +21,7 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
-// Background modes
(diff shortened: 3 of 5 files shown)


src/config.h
```diff
@@ -65,9 +65,17 @@ void config_add_loader(const char* section, config_loader loader);
...
  * @return false if text has invalid format
  */
 bool config_to_color(const char* text, argb_t* color);
+
+/**
+ * Convert text value to ARGB color which may be any alpha.
+ * @param text text to convert
+ * @param color output variable
+ * @return false if text has invalid format
+ */
```
artemsen, line 80: I don't think we need two similar functions to parse color.

src/viewer.c
```diff
@@ -57,12 +55,13 @@ static const char* scale_names[] = {
...
-    size_t frame;         ///< Index of the current frame
-    argb_t image_bkg;     ///< Image background mode/color
-    argb_t window_bkg;    ///< Window background mode/color
-    bool antialiasing;    ///< Anti-aliasing mode on/off
-    bool fixed;           ///< Fix image position
+    ssize_t img_x, img_y;  ///< Top left corner of the image
+    size_t frame;          ///< Index of the current frame
+    argb_t image_bkg;      ///< Image background color
+    argb_t window_bkg;     ///< Window background color
+    bool image_bkg_grid;   ///< Whether image background is grid
```
artemsen, line 62: Now we can set  both values: color and grid at same time, but they are mutually exclusive options.

artemsen (commented):
Thank you for the PR!
I am far away from my laptop, but I will return to this review in July.

src/viewer.c
an-prata: I figured this was the most effective way to convey the background preferences, now that every possible value of `argb_t` is used, even if the color becomes redundant in the grid case. Regardless, the acceptance of arbitrary alpha values requires the type's size to increase since none of the existing values of `argb_t` are available for non-color interpretation.

Could be replaced by a 64 bit type or another struct, which if preferable I would be happy to add to the PR.

src/config.h
an-prata: I may be mistaken about the cause, but if the `config_to_translucent_color` is used for all colors then I get some weird effects on the overlay's font. I imagine this is because the alpha value is used in `text.c` for special meanings (absence/presence of shadow), unfortunately mirroring the original behavior of giving an alpha of `0x00` when none is given would result in background colors not working correctly when given an opaque color, as it would translate to being transparent (alpha of 0x00), rather than opaque (alpha of 0xff) like in the new function.

Could easily be paramterized, but behavior must differ between parsing `#RRGGBB` and `#RRGGBBAA`. I didn't parameterize it in order to minimize the PR's changes.

artemsen:
Yeah, I forgot about that hacks with special values in color.
We need to get rid of them, but may be not in this PR.
Anyway, my vacation doesn't allow me to deep into the code in next two
weeks :)

src/config.c
```diff
@@ -303,3 +304,27 @@ bool config_to_color(const char* text, argb_t* color)
 
     return false;
 }
+
+bool config_to_translucent_color(const char* text, argb_t* color)
```
artemsen, line 308: I think we could merge `config_to_color` and `config_to_translucent_color`.
If alpha is not present - just set it as `0xff`.

artemsen (commented):
Sorry for delay, I am back =)

src/config.c
an-prata: Alpha is set to `0xff` when not present in `config_to_translucent_color`, the reason I made a new function here is because fonts render incorrectly when the alpha is not `0x00` like would be the case in `config_to_color`.

Seperate functions:
image

Merged with `0xff` alpha:
image

I would just make the default fill for no alpha `0x00`, but we have no way of distinguishing an intentionally transparent color from an opaque one in that case.

artemsen:
Could you squash these commits?
And add `Signed-off-by` to the commit message please.

PR #145, opened 2024-06-22
Add interface to create image context from memory

nicolasbrailo:
Exposing this function enables users hacking Swayimg to implement their own image-to-memory loader (eg from curl) while keeping the standard image loader from upstream.

### DIFF
--- a/src/image.c
+++ b/src/image.c
@@ -140,6 +140,11 @@ done:
     return ctx;
 }
 
+struct image* image_create_from_memory(const uint8_t* data, size_t size)
+{
+    return image_create(MEM_FILE_NAME, data, size);
+}
+
 void image_free(struct image* ctx)
 {
     if (ctx) {
--- a/src/image.h
+++ b/src/image.h
@@ -37,6 +37,9 @@ struct image_info {
 /** Name used for image, that is read from stdin through pipe. */
 #define STDIN_FILE_NAME "{STDIN}"
 
+/** Name used for image that is read directly from memory. */
+#define MEM_FILE_NAME "{MEM}"
+
 /**
  * Load image from file.
  * @param file path to the file to load
@@ -49,6 +52,12 @@ struct image* image_from_file(const char* file);
  * @return image context or NULL on errors
  */
 struct image* image_from_stdin(void);
+
+/**
+ * Load image from an arbitrary memory location.
+ * @return image context or NULL on errors
+ */
+struct image* image_create_from_memory(const uint8_t* data, size_t size);
 
 /**
  * Free image.


nicolasbrailo:
Using this one for my custom hack of swayimg - hope to upstream it so I need to sync less patches when pulling upstream.

artemsen:
1. All images are loaded from memory, even stdio is read into memory before
load. So, this interface is just a wrapper.
2. But this wrapper is not used anywhere, it is a dead code that we can't
check.
3. Even if it is unused, it creates several problems. For example, segfault
on reload, broken image list logic and so on.

If I understand right, this hack is needed just for another hack. Let's
join them in a standalone branch/repo :)

Why do you need to mix curl with swayimg? May be stdio is enough for this?

nicolasbrailo:
> 2. But this wrapper is not used anywhere, it is a dead code that we can't

Can also change this PR to simply remove the static keywrd to the implementation, if that's safer from your PoV.

> If I understand right, this hack is needed just for another hack.

Yes, for the traditional definition of hack, as in "to modify a program in a clever way" - not because either this patch or my branch are necessarily of low quality :)

> Why do you need to mix curl with swayimg? May be stdio is enough for this?

Not really, for my use case I need to have a set of images in memory, and be able to cycle through them in a ringbuffer. I also need to fetch them from a server. I understand curl is not a dep you'd like to upstream, and that's fine; as long as I can keep the curl changes within a single .c file or two, I can maintain my branch and periodically pull from upstream. If I need to maintain a set of patches, it's probably too much hassle to try to upstream changes, and I should just fork to keep it simple for both projects.

artemsen:
I have an idea.
What if I add support for external loaders? It will make it possible to specify a command to get image data, something like
```
$ swayimg "stdio://curl http://server/img.jpg" "stdio://downsample.sh ftp://server/img2.png"
```

nicolasbrailo:
> What if I add support for external loaders? 

Hard to say for sure if it'd work for my quite specific use case. It may :)
I'd, basically, need to invoke this loader following the slideshow rules on timeout (ie I just reload the same url, and get a different picture. Then I keep this in a ringbuffer, for history and prefetch). As long as it returns an image decoded to memory (and it's not an unreasonable amount of work for a niche fork!) it could work.

artemsen:
>Hard to say for sure if it'd work for my quite specific use case. It may :)

Yeah. It is definitely quite specific use case =)
I have implemented loading from external commands and cache configuration. May be this help.
And current version already has [image_from_memory](https://github.com/artemsen/swayimg/blob/ab676a6aa9790013057f4375d9a7cc8ee1374ef9/src/loader.c#L347), so I close this issue.

If you have any questions or ideas, please feel free to open new PR.

PR #146, opened 2024-06-22
[RFC] Downsample when decoding

nicolasbrailo:
When used in very constrained systems (eg raspberry pi's) the memory pressure from Swayimg can be quite significant. It's very easy to bring down a platform with 512mb of RAM just by loading a few large pictures and keeping them in memory. Implementing this change would enable loaders to reduce image size, so that the actual pixels stored in memory are not more than those that will be rendered to screen.

In my system, using not-very-large images (12mp) I see a reduction in memory usage of about 1/3, and faster render speeds.

### DIFF
--- a/src/formats/jpeg.c
+++ b/src/formats/jpeg.c
@@ -35,7 +35,7 @@ static void jpg_error_exit(j_common_ptr jpg)
 
 // JPEG loader implementation
 enum loader_status decode_jpeg(struct image* ctx, const uint8_t* data,
-                               size_t size)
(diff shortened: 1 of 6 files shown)


artemsen:
I don't think it is a good idea.
Only jpeg has this feature, it useless for other formats.

12 megapixels is 48 MiB of memory, not so much even for raspberry. We could
disable cache and turn off preload functionality, this will save 2/3 of RAM.

artemsen:
Sorry, I can't merge this code as it solves a very specific and narrowly focused problem.

PR #149, opened 2024-07-09
Add no-operation action for key bindings

g-rden:
I didn't want "Key Super_L is not bound" to be displayed when using Super_L do control my WM. I think this could be useful for other people's setups as well.


artemsen:
In fact we don't need to create one more dummy type, it is enough to add Super to the exception list.
But it is possible now to suppress any key with existing "none" action.

Thank you for PR!

PR #151, opened 2024-07-10
Fix font height offsets

### LINKED ISSUE #139: Incorrect rendering of the character i
tmp RNqi1LeSqa

Every `i` reaches below its line. This happens with both Inter and JetBrainsMono.

g-rden:
The allocated line height is now bigger, because the glyphs didn't fit in it. The changes in text.c are just for the line height to be similar to the old one.

Comparison with fira 9px:
left: new, right: old
new-old

Closes #139

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -13,6 +13,7 @@
 #include FT_FREETYPE_H
 #include FT_GLYPH_H
 
+#define POINT_FACTOR 64 // default points per pixel for 26.6 format
 #define SPACE_WH_REL 2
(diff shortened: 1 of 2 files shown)


artemsen:
Wow!!! Thank you!
New version has cropped characters, but it definitely looks better than jumped glyphs.

Play font:
image

g-rden:
Missed that entirely. Liberation font is also cropped. I'm not sure how to deal with this.

PR #154, opened 2024-07-11
Fix color blending and masking

g-rden:
Fixes that text could never be fully transparent, edges of text wouldn't be more transparent than the rest of the text (pixelated) and makes colors blend as expected.

Upper one is current, lower one is this patch:
blend-mask-fix
When checking the color values in gimp or something, the new one is correct.


g-rden:
Just noticed I should have used 0xff instead of 255. Sorry

artemsen:
Don't' worry, what matters is that it works =)

PR #157, opened 2024-07-15
Disable antialiasing while dragging

g-rden:
While the left mouse button is pressed down antialiasing is now disabled.

Reason: Dragging while antialiasing is enabled is unusable on some machines.

### DIFF
--- a/src/ui.c
+++ b/src/ui.c
@@ -347,6 +347,7 @@ static void on_pointer_button(void* data, struct wl_pointer* wl_pointer,
     if (button == BTN_LEFT) {
         ctx.mouse.active = (state == WL_POINTER_BUTTON_STATE_PRESSED);
     }
+    ui_redraw();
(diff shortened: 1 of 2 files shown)


artemsen:
This is a very extraordinary solution, but its implementation looks like a hack, especially "let's redraw on every mouse key event" =)
The main complaint is that the entire solution is based on the fact that `ui_redraw()` is a synchronous operation. This works now, but I plan to change the logic to event-driven model and ui_redraw will be called asynchronously.
So, even if we merge this PR, this solution will not survive to release.

g-rden:
> very extraordinary solution

lol, yeah I am no programmer.

> "let's redraw on every mouse key event"

That is not happening. It only gets drawn when the mouse button is first pressed down and then when it is released again.

> this solution will not survive to release.

Fair.

PR #158, opened 2024-07-16
Fix 1.0 scale

g-rden:
Previously pixman_copy was used when the image was scaled 1 to 1. Easiest way to see what's wrong with that is when viewing an image with the resolution of the display in fullscreen and moving the image up or left.


artemsen:
In fact, it was a bug in `pixmap_copy()`

PR #161, opened 2024-07-24
Unbreak build on FreeBSD

jbeich:
Looks like CI can be green if building the app fails but tests succeed.


artemsen:
> Looks like CI can be green if building the app fails but tests succeed.

It turned out that it was always green =) Fixed.

PR #172, opened 2024-08-04
Sanitizer

appashchenko:
Fix various problems found by sanitizer.

### DIFF
--- a/src/application.c
+++ b/src/application.c
@@ -380,9 +380,8 @@ bool app_init(const char** sources, size_t num)
     }
     if (ctx.window.width == SIZE_FULLSCREEN) {
         ui_toggle_fullscreen();
-    } else if (first_image &&
-               (ctx.window.width == SIZE_FROM_IMAGE ||
-                ctx.window.width == SIZE_FROM_PARENT)) {
+    } else if (ctx.window.width == SIZE_FROM_IMAGE ||
+               ctx.window.width == SIZE_FROM_PARENT) {
         // fixup window size form the first image
         const struct pixmap* pm = &first_image->frames[0].pm;
         ctx.window.width = pm->width;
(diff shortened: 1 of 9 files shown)


src/application.c
```diff
@@ -380,9 +380,8 @@ bool app_init(const char** sources, size_t num)
     }
     if (ctx.window.width == SIZE_FULLSCREEN) {
         ui_toggle_fullscreen();
-    } else if (first_image &&
```
artemsen, line 383: `first_image` always exists, see line 373.

src/config.c
```diff
@@ -312,7 +316,7 @@ bool config_to_color(const char* text, argb_t* color)
     if (strlen(text) > 6) { // value with alpha (RRGGBBAA)
         *color = (num >> 8) | ARGB_SET_A(num);
     } else {
-        *color = num | ARGB_SET_A(0xff);
+        *color = num | ARGB_SET_A(0xffu);
```
artemsen, line 319: It makes the code less readable. Please, remove these `u`s.

src/application.c
appashchenko: > `first_image` always exists, see line 373.

`first_image` can not be NULL after line 374 so it is always true at line 383.

artemsen: Exactly! So, we don't need to check it twice =)

appashchenko: Артём, вы не выспались, или я что-то не понимаю?
Коммит наоборот **убирает** эту переменную из условия.

artemsen:
Видимо не выспался =) Да, тут всё норм.

src/config.c
appashchenko: You are right. it was bad decision.
I have reverted this part and replaced 'u' with cast to argb_t where is needed.
Please review changed commit.

PR #182, opened 2024-08-27
Fix infinite loop when going to the first/last image and it is invalid

juodumas:


artemsen:
Good catch!
Thank you!

PR #193, opened 2024-10-09
Add QOI format support

sepisoad:
this PR adds support to QOI image format: https://qoiformat.org/
for sake of simplicity I have included the official QOI encoder/decoder in the source so there is no need for external dependency


sepisoad:
recreated my changes in another PR https://github.com/artemsen/swayimg/pull/195

PR #195, opened 2024-10-09
Add QOI Image format support

sepisoad:
this PR adds support to QOI image format: https://qoiformat.org/
for sake of simplicity I have included the official QOI encoder/decoder in the source so there is no need for external dependency


sepisoad:
i can't figure why this fails: https://github.com/artemsen/swayimg/actions/runs/11261487331/job/31315036603?pr=195#step:7:79

appashchenko:
> i can't figure why this fails: https://github.com/artemsen/swayimg/actions/runs/11261487331/job/31315036603?pr=195#step:7:79

Yout should enable qoi loader in `test/loader_test.cpp` and `test/meson.build`
Create new **2x2** pixels qoi image at `test/data/image.qoi`

appashchenko:
Honestly I do not like how it looking.
The decoder can be [more adopted](https://gist.github.com/appashchenko/ab1b59b4965a17e987908b304438a1d5).

sepisoad:
@appashchenko i'm not sure if i understand what you mean. do you mean that we should only keep the decoder part of the header file and incorporate it into the qoi.c file?

appashchenko:
@sepisoad 
1) yes, decoder is a single .c file;
2) you do not need allocate a pixel buffer and copy memory when you already have an allocated frame and its pixmap. Use this;
3) a lot of useless for `swayimg` defines and a LOT of comments. Code should be clean;

Anyway thanks for your work.

There is an another pull request for QOI #196

PR #201, opened 2024-10-18
feat: storing thumbnails on disk

### LINKED ISSUE #200: feat: storing thumbnails on disk
I don't know if this is completely outside the scope of the project, but I did a quick and dirty hack that caches thumbnails on disk similarly to how nsxiv does it. I ran some benchmarks and it saves 10 to 20 ms per image. The gallery still takes a very long time to load but the difference is there. If this is something that might get merged, I'd be happy to make a PR.

Rentib:
Closes #200


src/image.c
```diff
@@ -65,6 +66,10 @@ void image_thumbnail(struct image* image, size_t size, bool fill,
     size_t thumb_height = scale * full->height;
     ssize_t offset_x, offset_y;
     enum pixmap_scale scaler;
+    static char* cache_dir = NULL;
+    char thumb_path[4096] = { 0 }; /* PATH_MAX */
```
artemsen, line 70: `PATH_MAX` is defined in `limits.h`, we can use it instead of magic numbers.

src/image.c
```diff
@@ -82,12 +87,53 @@ void image_thumbnail(struct image* image, size_t size, bool fill,
         offset_y = 0;
     }
 
+    // TODO: probably should be moved outside of this function
```
artemsen, line 90: I agree =)

src/image.c
```diff
@@ -82,12 +87,53 @@ void image_thumbnail(struct image* image, size_t size, bool fill,
...
+    if (!cache_dir) {
+        cache_dir = getenv("XDG_CACHE_HOME");
+        if (cache_dir) {
+            cache_dir = strcat(cache_dir, "/swayimg");
+        } else {
+            cache_dir = getenv("HOME");
+            if (!cache_dir) {
+                goto thumb_create;
+            }
+            cache_dir = strcat(cache_dir, "/.swayimg");
```
artemsen, line 100: `HOME/.cache/swayimg`, but it would be better to have an option in the config file so the user could set it themselves.

src/image.c
```diff
@@ -82,12 +87,53 @@ void image_thumbnail(struct image* image, size_t size, bool fill,
...
+        }
+    }
+    sprintf(thumb_path, "%s%s", cache_dir, image->source);
+
+    if (!pixmap_load(&thumb, thumb_path)) {
+        goto thumb_create;
+    }
+
+    if (!(thumb.width == thumb_width && thumb.height == thumb_height)) {
```
artemsen, line 109: What if the file has been changed but the image size remains the same?
In this case, we need to recreate the thumbnail.

src/image.c
```diff
@@ -82,12 +87,53 @@ void image_thumbnail(struct image* image, size_t size, bool fill,
...
     // create thumbnail
     if (!pixmap_create(&thumb, thumb_width, thumb_height)) {
         return;
     }
     pixmap_scale(scaler, full, &thumb, offset_x, offset_y, scale, image->alpha);
 
+    // save thumbnail to disk
+
+    // TODO: probably should be moved outside of this function
+    // TODO: do not use system() here
```
artemsen, line 124: Definitely.

src/pixmap.c
```diff
@@ -300,6 +301,70 @@ void pixmap_free(struct pixmap* pm)
...
+bool pixmap_save(struct pixmap* pm, const char* path)
+{
+    FILE* fp;
+    unsigned i;
+
+    if (!(fp = fopen(path, "wb"))) {
+        return false;
+    }
+
+    // TODO: add alpha channel
```
artemsen, line 313: We have 14 supported image formats. It would be better to use one of them instead of reinventing own format.

artemsen (changes requested):
I think this feature should be optional. We need an option in the config file that will allow saving thumbnails to disk. Writing anything to persistent storage by default is not a good idea.

src/image.c
Rentib: Adding this to config would be better for another PR i think, but I can try adding it here too...

src/image.c
Rentib: I have no idea how to check this to be honest. Some libnotify thing?

src/pixmap.c
Rentib: I don't see any function for writing an image to disk in the codebase. Did I miss something?

src/image.c
artemsen: Yes, this can be taken out into a separate request.
But in any case, this PR should contain the configurable option for enabling/disabling the persistent cache.

src/image.c
artemsen: I would compare mtime of the files.

src/pixmap.c
artemsen: Current code doesn't contain code for encoders. But these APIs are available to `swayimg`.

src/image.c
```diff
@@ -4,9 +4,13 @@
 
 #include "image.h"
 
+#include <errno.h>
+#include <linux/limits.h>
```
artemsen, line 8: This breaks FreeBSD build.

src/image.c
```diff
@@ -51,6 +55,63 @@ void image_rotate(struct image* ctx, size_t angle)
...
+    return r >= 0 && r < PATH_MAX;
+}
+
+// TODO: where should this function be?
+/**
+ * Makes directories like `mkdir -p`.
+ * @param path absolute path to the directory to be created
+ * @return true if successful
+ */
+static bool make_directories(char* path)
```
artemsen, line 91: `make_directory` in image entity. Something went wrong in program architecture =)

src/pixmap.c
```diff
@@ -300,6 +301,70 @@ void pixmap_free(struct pixmap* pm)
     free(pm->data);
 }
 
+bool pixmap_save(struct pixmap* pm, const char* path)
+{
+    FILE* fp;
+    unsigned i;
```
artemsen, line 307: Please, use fixed-size types.

src/image.c
```diff
@@ -51,6 +55,63 @@ void image_rotate(struct image* ctx, size_t angle)
...
+{
+    char* slash;
+
+    if (!*path) {
+        return true;
```
artemsen, line 96: Why this function may be called for empty directory?

src/image.c
```diff
@@ -51,6 +55,63 @@ void image_rotate(struct image* ctx, size_t angle)
...
+
+    slash = strchr(slash + 1, '/');
+    while (slash) {
+        *slash = '\0';
+        if (mkdir(path, 0755) && errno != EEXIST) {
+            // TODO: do we want to print error message?
+            return false;
+        }
+        *slash = '/';
+        slash = strchr(slash + 1, '/');
```
artemsen, line 109: Copy/paste from line 101.

src/image.c
Rentib: Then mtime would have to be stored somehow in the thumbnail or another file with mtimes of cached images.
Is there some better method for this?

src/image.c
```diff
@@ -51,6 +55,63 @@ void image_rotate(struct image* ctx, size_t angle)
...
+
+    if (!*path) {
+        return true;
+    }
+
+    slash = path;
+
+    slash = strchr(slash + 1, '/');
+    while (slash) {
+        *slash = '\0';
```
artemsen, line 103: It's a bad practice. You modify the input buffer, that should be const.

src/image.c
```diff
@@ -82,12 +144,25 @@ void image_thumbnail(struct image* image, size_t size, bool fill,
         offset_y = 0;
     }
 
+    // get thumbnail from cache
+    if (get_image_thumb_path(image, thumb_path) &&
```
artemsen, line 148: `image_thumbnail` is intended to create a thumbnail, not load.
Any function should do one thing =)

src/image.c
```diff
@@ -51,6 +55,63 @@ void image_rotate(struct image* ctx, size_t angle)
...
+            cache_dir = strcat(cache_dir, "/swayimg");
+        } else {
+            cache_dir = getenv("HOME");
+            if (!cache_dir) {
+                return false;
+            }
+            cache_dir = strcat(cache_dir, "/.swayimg");
+        }
+    }
+    r = snprintf(path, PATH_MAX, "%s%s", cache_dir, image->source);
```
artemsen, line 81: `snprintf(path, sizeof(path), ...`

src/image.c
```diff
@@ -51,6 +55,63 @@ void image_rotate(struct image* ctx, size_t angle)
...
+ * @return true if successful
+ */
+static bool get_image_thumb_path(struct image* image, char* path)
+{
+    static char* cache_dir = NULL;
+    int r;
+    if (!cache_dir) {
+        cache_dir = getenv("XDG_CACHE_HOME");
+        if (cache_dir) {
+            cache_dir = strcat(cache_dir, "/swayimg");
```
artemsen, line 72: Wow =)

src/pixmap.c
```diff
@@ -300,6 +301,70 @@ void pixmap_free(struct pixmap* pm)
...
+        return false;
+    }
+
+    char header[3];
+    if (fscanf(fp, "%2s\n%zu %zu\n255\n", header, &pm->width, &pm->height) !=
+        3) {
+        goto fail;
+    }
+
+    if (strcmp(header, "P6")) {
```
artemsen, line 341: Just curious, what does "P6" mean?

src/pixmap.c
```diff
@@ -300,6 +301,70 @@ void pixmap_free(struct pixmap* pm)
...
+    unsigned i;
+
+    if (!(fp = fopen(path, "wb"))) {
+        return false;
+    }
+
+    // TODO: add alpha channel
+    fprintf(fp, "P6\n%zu %zu\n255\n", pm->width, pm->height);
+    for (i = 0; i < pm->width * pm->height; ++i) {
+        uint8_t color[4] = { (((pm->data[i] >> (8 * 2)) & 0xff)),
```
artemsen, line 316: 4-items array, but filled only 3.

artemsen (changes requested):

src/image.c
artemsen: Why not use the file system for this? That's what it's designed for =)

Rentib: Ah, I get it. Compare the mtime of thumbnail and image and if image's mtime is greater than remake the thumbnail, right?

src/image.c
Rentib: I don't really see anything wrong with it, but I changed it.

src/image.c
Rentib: I don't know where to put it...

src/image.c
Rentib: Should I change gallery.c: add_thumbnail() to first try to get thumbnail from cache and if it is unavailable, then run image_thumbnail() and store it on disk?

src/image.c
Rentib: It must be ctime

src/image.c
artemsen: I would create `thumbnail.c` and move all these changes there (include pixman).

src/pixmap.c
artemsen: No, [it's not](https://github.com/artemsen/swayimg/blob/master/src/formats/bmp.c).
But you can add real BMP encoder for that =).

src/image.c
artemsen: One more thing: if antialising mode or fill mode were changed, then we have to recreate thumbnails too.

Rentib: This makes sense, do you have any idea on how to check this?
I could maybe look at config's ctime and recreate each time it is changed, but I don't really like this idea.

artemsen: Store this data within the thumbnail? In meta/exif/...

src/pixmap.c
Rentib: Ah sorry, my mistake. It's PPM format. Probably the simplest thing out there.

src/image.c
Rentib: done

src/image.c
Rentib: Would it be fine to put `export_path()` in config.h?
Then I could use it here

src/pixmap.c
artemsen: I think it would be better to add `encode_pnm` to [PNM(PPM)](https://github.com/artemsen/swayimg/blob/master/src/formats/pnm.c) decoder.

artemsen:
Please, rebase to latest master. This will fix Ubuntu CI.

src/image.c
Rentib: Sorry, I meant `expand_path()`.

artemsen: This function is specific to thumbnails, so leave it in `thumbnail.c`.

Rentib: Let me rephrase the question:
Would it be alright to add `expand_path()` to config.h header file?
It is currently implemented as a static function in config.c:203

artemsen: Ah, I get it =)
Yes, it will be ok.

src/thumbnail.c
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    if (!(fp = fopen(path, "rb"))) {
+        goto fail;
+    }
+
+    char header[3];
+    if (fscanf(fp, "%2s\n%zu %zu\n255\n", header, &thumb->width,
+               &thumb->height) != 3) {
+        goto fail;
```
artemsen, line 179: You love `goto` so much =)
Maybe there is another way to handle errors?

artemsen (changes requested):

src/thumbnail.c
```diff
@@ -0,0 +1,242 @@
...
+#elif defined(__OpenBSD__) || defined(__FreeBSD__)
+#define PATH_MAX 4096
+#endif
+
+/**
+ * Makes directories like `mkdir -p`.
+ * @param path absolute path to the directory to be created
+ * @return true if successful
+ */
+static bool make_directories(char* path)
```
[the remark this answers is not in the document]
Rentib: Then I would need to strdup it

src/thumbnail.c
```diff
@@ -0,0 +1,242 @@
...
+{
+    char* slash;
+
+    if (!path || !*path) {
```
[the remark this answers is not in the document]
Rentib: Yes, but isn't it better to be safe? I can remove the condition, but it doesn't really impact the performance.

src/thumbnail.c
```diff
@@ -0,0 +1,242 @@
...
+        if (mkdir(path, 0755) && errno != EEXIST) {
+            return false;
+        }
+        *slash = '/';
+    }
+
+    return true;
+}
+
+static bool get_thumb_path(char** path, const char* source)
```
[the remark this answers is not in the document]
Rentib: Ok, this is a problem. I will fix it to not pass source, but the absolute file path.

src/thumbnail.h
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: MIT
+// Image instance: pixel data and meta info.
+// Copyright (C) 2021 Artem Senichev <artemsen@gmail.com>
```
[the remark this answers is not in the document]
Rentib: Function `thumbnail_create()` is your work (I copy/pasted). I can remove the copyright if you want though.

src/thumbnail.c
Rentib: This is the same way of handling errors as in `formats/avif.c` and many other files in the project. I could maybe make a macro which cleans stuff up and then returns false, but I believe this is cleaner.

src/thumbnail.c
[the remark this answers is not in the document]
artemsen: It is not about performance, but about architecture. Why do we have cases when path is NULL or empty?

src/thumbnail.c
```diff
@@ -0,0 +1,242 @@
...
+{
+    char* slash;
+
+    if (!path || !*path) {
+        return false;
+    }
+
+    slash = path;
+    while (true) {
+        slash = strchr(slash + 1, '/');
```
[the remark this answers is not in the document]
Rentib: I don't see an easy way to concatenate slices. I need to mkdir a prefix of `path` every time.

artemsen:
@Rentib, I appreciate your efforts and thanks for the prototype!

PR #207, opened 2024-10-26
feat(#205): add default/global scale action

### LINKED ISSUE #205: Add a command for changing the `scale`
Currently, you can change the scale only for the current image using `zoom <scale mode>`. I would really like to be able to set the scaling mode for all images, like it is being set when starting the program -> `scale <scale mode>`. Perhaps even cycle through the modes like with `mode` or `antialiasing`.

I would use `scale real` often to compare the sharpness of two similar images (as I do now on _imv_).

Thank you for this great application!

litoj:
Enables users to change the default scaling mode set in config also during runtime.

Resolves #205


.editorconfig
```diff
@@ -0,0 +1,6 @@
+root=true
```
artemsen, line 1: Remove this file, please.

src/viewer.c
```diff
@@ -297,6 +295,33 @@ static void zoom_image(const char* params)
...
+        if (fixed_scale >= 0) {
+            ctx.scale_init = fixed_scale;
+        } else {   
+            fprintf(stderr, "Invalid scale operation: \"%s\"\n", params);
+            return;
+        }
+    } else {
+        // toggle to the next scale
+        ctx.scale_init++;
+        if (ctx.scale_init >= sizeof(scale_names) / sizeof(char*)) {
```
artemsen, line 316:
```c
if (ctx.scale_init >= ARRAY_SIZE(scale_names)) {
```

artemsen (changes requested):

artemsen:
Thank you!

PR #215, opened 2024-11-18
Fix cache indexing out-of-bounds

Player01osu:
Next image and previous image commands in viewer mode segfaults when insufficient amount of images exist (specifically when `size_t size` in `cache_trim` is less than `cache->capacity - 1`).


artemsen:
Thank you!

PR #216, opened 2024-11-19
Added farbfeld parser

alttabber:
Added support for the [farbfeld image format](https://tools.suckless.org/farbfeld/).

Since farbfeld is suppoed to be externally compressed (`foo.ff.gz`) another PR is supposed to follow to support single file external compression.


src/formats/ff.c
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: MIT
+// farbfeld format decoder
+
+#include "../loader.h"
+
+#include <limits.h>
```
artemsen, line 6: Unused header.

src/formats/ff.c
```diff
@@ -0,0 +1,40 @@
...
+#include <stdint.h>
```
artemsen, line 7: This header is already included by "../loader.h".

src/formats/ff.c
```diff
@@ -0,0 +1,40 @@
...
+{
+    size_t width, height;
+
+    if (size < 16 ) {
+        return ldr_unsupported;
+    }
+
+    uint8_t mByte[8] = {'f', 'a', 'r', 'b', 'f', 'e', 'l', 'd'};
+
+    for (int i = 0; i < 8; i++) if (data[i] != mByte[i]) return ldr_unsupported;
```
artemsen, line 21: Can be simplified with `memcmp`. See other decoder for example.

src/formats/ff.c
```diff
@@ -0,0 +1,40 @@
...
+enum loader_status decode_ff(struct image* ctx, const uint8_t* data,
+                              size_t size)
+{
+    size_t width, height;
+
+    if (size < 16 ) {
+        return ldr_unsupported;
+    }
+
+    uint8_t mByte[8] = {'f', 'a', 'r', 'b', 'f', 'e', 'l', 'd'};
```
artemsen, line 19: 1. Should be `const`;
2. Use snake_case for variables.

src/formats/ff.c
```diff
@@ -0,0 +1,40 @@
...
+
+    ctx->alpha=true;
+
+    if (!image_allocate_frame(ctx, width, height)) return ldr_fmterror;
+
+    struct pixmap* pm = &ctx->frames[0].pm;
+
+    for (size_t i = 16; i < size; i=i+8) {
+        argb_t pixel = ARGB(data[i+6], data[i], data[i+2], data[i+4]);
+	pm->data[(i-16)/8] = pixel;
```
artemsen, line 34: Segfault if file is corrupted, e.g. `width * height * sizeof(argb) < size`.

src/formats/ff.c
```diff
@@ -0,0 +1,40 @@
...
+
+    if (size < 16 ) {
+        return ldr_unsupported;
+    }
+
+    uint8_t mByte[8] = {'f', 'a', 'r', 'b', 'f', 'e', 'l', 'd'};
+
+    for (int i = 0; i < 8; i++) if (data[i] != mByte[i]) return ldr_unsupported;
+
+    width = data[11] + data[10]*256 + data[9]*65536 + data[8]*1638400;
```
artemsen, line 23: Please use clang-format, the config is in the project's root dir.

src/loader.c
```diff
@@ -134,7 +135,7 @@ static const image_decoder decoders[] = {
 #ifdef HAVE_LIBTIFF
     &LOADER_FUNCTION(tiff),
 #endif
-    &LOADER_FUNCTION(qoi),  &LOADER_FUNCTION(tga),
+    &LOADER_FUNCTION(qoi),  &LOADER_FUNCTION(tga), &LOADER_FUNCTION(ff)
```
artemsen, line 138: Please move ff decoder in front of tga (tga doesn't have any magic id and should be the last decoder).

src/formats/ff.c
```diff
@@ -0,0 +1,40 @@
...
+    height = data[15] + data[14]*256 + data[13]*65536 + data[12]*1638400;
```
artemsen, line 24: You can use one of `bswap_32/be32toh/htonl` instead this strange formula.

artemsen (changes requested):

src/formats/ff.c
artemsen: Oh.... Let's simulate =)
```
with = 10;
height = 10;
size = 10000000000;
```

alttabber: I have tried both exceeding size (`size > 16+width*height*sizeof(argb_t)*2`) and insufficient size (`size < 16+width*height*sizeof(argb_t)*2`)

In the case of insufficient size, as expected, fails the line 27 test. In case of exceeding size, as expected, displays garbage. No segfault in either case.

The exceeding size is matter of policy: it's better to display a broken image or to just fail the decode?

src/formats/ff.c
```diff
@@ -0,0 +1,43 @@
...
+
+    if (memcmp(data, magic, 8) != 0) return ldr_unsupported;
+
+    memcpy(&width, (const uint8_t*) data+8, 4);
+    width = htonl(width);
+    memcpy(&height, (const uint8_t*) data+12, 4);
+    height = htonl(height);
+
+    // 16 bytes of metadata, plus farbfeld is exclusively a 16 bit format
+    if (16+width*height*sizeof(argb_t)*2 > size) return ldr_fmterror;
```
artemsen, line 27: By the way, why `*2`?

src/formats/ff.c
```diff
@@ -0,0 +1,43 @@
...
+
+    if (size < 16 ) {
+        return ldr_unsupported;
+    }
+
+    static uint8_t magic[8] = {'f', 'a', 'r', 'b', 'f', 'e', 'l', 'd'};
+
+    if (memcmp(data, magic, 8) != 0) return ldr_unsupported;
+
+    memcpy(&width, (const uint8_t*) data+8, 4);
```
artemsen, line 21: It's not necessary, `width = *(const uint32_t*) & data[8]`.
But it would be better i if you define a simple structure something like this:
```
struct farbfeld_header {
    uint8_t magic[8];
    uint32_t width;
    uint32_t height;
};
```

src/formats/ff.c
```diff
@@ -0,0 +1,43 @@
...
+enum loader_status decode_ff(struct image* ctx, const uint8_t* data,
+                              size_t size)
+{
+    size_t width, height;
+
+    if (size < 16 ) {
+        return ldr_unsupported;
+    }
+
+    static uint8_t magic[8] = {'f', 'a', 'r', 'b', 'f', 'e', 'l', 'd'};
```
artemsen, line 17: const

artemsen (changes requested):

src/formats/ff.c
alttabber: argb_t assumes is 8 bit per channel, farbfeld is 16.

src/formats/ff.c
artemsen: It's better to display broken image.

src/formats/ff.c
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+// farbfeld format decoder
+
+#include "../loader.h"
+#include <netinet/in.h>
```
artemsen, line 5: [#include <arpa/inet.h>](https://linux.die.net/man/3/htonl)

src/formats/ff.c
alttabber: farbfeld *only* supports 16 bit BE, and is assumed sRGB straight alpha. Since the image is supposed to be sRGB, I can just cut off the less significant bits and still have an sRGB image, just a less precise one.

src/formats/ff.c
```diff
@@ -0,0 +1,64 @@
...
+    // 16 bytes of metadata, plus farbfeld is exclusively a 16 bit
+    // format but argb_t assumes 8 bit
+    return sizeof(struct farbfeld_header) +
+        header.width * header.height * sizeof(argb_t) * 2;
+}
+
+enum loader_status decode_ff(struct image* ctx, const uint8_t* data,
+                             size_t size)
+{
+    struct farbfeld_header header;
```
artemsen, line 26: `const struct farbfeld_header* header = data;` and so on

src/formats/ff.c
```diff
@@ -0,0 +1,64 @@
...
+
+    if (size < sizeof(struct farbfeld_header)) {
+        return ldr_unsupported;
+    }
+
+    const uint8_t magic[8] = { 'f', 'a', 'r', 'b', 'f', 'e', 'l', 'd' };
+
+    if (memcmp(data, magic, 8) != 0) {
```
artemsen, line 34: Please get rid of magic numbers.

src/formats/ff.c
```diff
@@ -0,0 +1,64 @@
...
+    ctx->alpha = true;
+
+    if (!image_allocate_frame(ctx, header.width, header.height)) {
+        return ldr_fmterror;
+    }
+
+    struct pixmap* pm = &ctx->frames[0].pm;
+
+    for (size_t i = sizeof(struct farbfeld_header);
+         i < min(size, ff_size(header)); i = i + 8) {
```
artemsen, line 56: `min` is called on each iteration.

src/formats/ff.c
```diff
@@ -0,0 +1,64 @@
...
+        argb_t pixel = ARGB(data[i + 6], data[i], data[i + 2], data[i + 4]);
+        pm->data[(i - sizeof(struct farbfeld_header)) / 8] = pixel;
```
artemsen, line 58: Magic `8`

artemsen (changes requested):

PR #217, opened 2024-11-19
Support single file external compression

alttabber:
Support single file external compression (ex. `foo.pnm.gz`, `foo.png.xz`.

This can be done by adding a recursive step to the decoder logic. Reference implementation using bzip2.

Related to #216

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -55,6 +55,9 @@ tiff = dependency('libtiff-4', required: get_option('tiff'))
 webp = dependency('libwebp', required: get_option('webp'))
 webp_demux = dependency('libwebpdemux', required: get_option('webp'))
 
+# optional dependencies: compression formats support
+bzip2 = dependency('bzip2', required: get_option('bzip2'))
+
 # hack to build "pkgconfigless" gif in FreeBSD
 gif_opt = get_option('gif')
 if gif_opt.disabled()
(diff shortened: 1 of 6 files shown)


alttabber:
In the case of farbfeld, expternal compression is basically part of the spec. Might as well go a step further and implement it for every format.

artemsen:
> A compressed file isn't necessarily an archive, it's foo.gz, not foo.tar.gz

It doesn't matter, just use it `gunzip ... | swayimg -`

> In the case of farbfeld, expternal compression is basically part of the spec. 

Then bzip should be a part of the farbfeld decoder. But I wouldn't want to get another dependency for the sake of something that can be done using standard console tools.

> Might as well go a step further and implement it for every format

Most image formats are already compressed, so (for example) using zip for a png that is already zipped is a waste of time.

alttabber:
The thing about using pipes is that you can only pipe one thing at a time. It's fine for batch processing, but not for anything that opens a window.

Let's say I don't want to open a single image, but an entire directory of mixed format images, maybe even recursively. I'd have to write a script that replaces the simple `.` argument with actually recursively navigating the file system, and replace all the compressed files with `"exec://$(decompressor) $(filename)"`. Not exactly the trivial task I want to do on the fly.

It might be quite the task, but how about an extension-based autoexec? It would go in the swayimgrc, and it would look like this:
```
[autoexec]
bz2=bcat %
```
And every time there's a `foo.bz2` it reads the stdout of `bcat foo.bz2`.

It would solve the use case without requiring swayimg to explicitly speak compression formats, while also significantly improving extensibility. The issue might be the difficulty of implementation.

artemsen:
> Not exactly the trivial task I want to do on the fly.

```
find . -name '*.gz' | sed 's,^,"exec://gunzip -c ",' | xargs swayimg
```
You can replace gunzip with unrar/7z/... and get support for other compression methods.
We don't need to write C code for this.

And once again, almost all modern formats are already compressed, there is no point in compressing them again.
Therefore, I think that adding bzip support to swayimg is not necessary.

alttabber:
You forgot the "mixed format", that's the part that makes it non-trivial.

An opinion on the autoexec proposal? Are you interested in at least a PR?

alttabber:
Superseded by #219

PR #219, opened 2024-11-20
Automatic generation of exec entries based of file exntension

alttabber:
This allows the user to set an exec rule that would be executed automatically when a file with a specified extension is opened by Swayimg.

Ex.
```
[listl.autoexec]
formats = mp4,eps
mp4 = ffmpegthumbnailer -i % -o - -c png
eps = inkscape % --export-filename=- --export-type=svg
```

In this example, every time the user tries to open a `foo.mp4` time, swayimg will automatically invoke ffmpegthumbnailer, and every time the user tries to open a `foo.eps` file, swayimg will automatically invoke Inkscape to convert it to svg.

This may have uses beyond just hacking in support for formats swayimg has no decoder for.

### DIFF
--- a/src/imagelist.c
+++ b/src/imagelist.c
@@ -11,6 +11,7 @@
 #include <string.h>
 #include <sys/stat.h>
 #include <time.h>
+#include <stdio.h>
 
 // Default configuration parameters
 #define CFG_ORDER_DEF     "alpha"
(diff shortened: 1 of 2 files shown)


artemsen:
The first point of [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) is:
Make each program do one thing well. To do a new job, build afresh rather than complicate old programs by adding new "features".

Please stop creating an all-in-one combine from an image viewer.
Image viewer should display images, that's all.
It shouldn't extract images from archives or resource files, it shouldn't convert them from one format to another, it shouldn't work directly with network or devices, it shouldn't do anything except decoding and displaying image data.

alttabber:
This PR doesn't add any of the features you listed. All it does is adding a UI to the already existing `exec://` that's in the config rather than in the CLI.

So, either `exec://` breaks the Unix philosophy already, or this PR doesn't.

artemsen:
Sorry, I still think batch processing is beyond the functionality of the image viewer and should be done by external tools.

Issue #225, opened 2024-11-28
Alpha blending


abewieland:
I don't think the current alpha blending function is quite right - it works fairly well when the background is fully opaque (the most common case in the code, when rendering to a window), but has some issues in any other cases. For example, if the background is fully transparent (for example when a new pixmap is created, for thumbnails or similar), we should expect to see the foreground pixels pass through as they are; in the current implementation that doesn't happen, however. Instead, the foreground colors are all multiplied by the alpha value, then never re-divided, so they end up darker than they should be. Consider the following image (entirely RGBA #00ff0080):
![001](https://github.com/user-attachments/assets/f0fce552-a282-4247-98eb-87bedd0ad4b6)

If we view it in swayimg over a black background (`viewer.transparency=#000000ff`), we'd expect to see a uniform #008000 color, and we do (I checked in gimp with a color picker):
![20241128_16h04m41s_grim](https://github.com/user-attachments/assets/0a3bcb89-42c6-47af-be36-f46b73c37e42)

We'd expect the thumbnail to be the same color, but it's darker in this case, because it's created as #00800080 (due to improper mixing), then rendered over black, giving us this (when all the gallery colors are all #000000ff, with no border or shadow):
![20241128_16h12m16s_grim](https://github.com/user-attachments/assets/fb97a677-5f79-410b-9d93-2f96b062d22d)
Here the color is #004000, which is darker than the actual image.

Another problem stems from APNGs, which may have frames which display over previous frames, like in this example: 
![in](https://github.com/user-attachments/assets/b78d0fe4-d904-423e-ba25-b3541a86890e)

We expect to see (and my browser has already correctly displayed) a bright green left half by the end of the animation, since the alpha should continue to rise on that side. In `swayimg`, however, both stay very dim (and the right is somehow grayish) since we're using `max`:
![20241128_16h18m49s_grim](https://github.com/user-attachments/assets/cdaef215-272f-4fac-a6b9-e540ecfc060c)
(here I used `viewer.transparency=#ffffffff`)

artemsen:
Fixed in #226

PR #226, opened 2024-11-28
Fix alpha blending

abewieland:
I used the formula from [here](https://en.wikipedia.org/wiki/Alpha_compositing#Description). That gave more consistent results, like the viewer and gallery agreeing on color:
20241128_16h38m27s_grim
20241128_16h40m59s_grim

And the APNG matches the browser a lot better, too:
20241128_16h37m09s_grim

Of course, we're doing blending in the sRBG space (rather than the linear space), so I didn't check whether it matches the browser exactly, but it's at least a lot closer.


abewieland:
Should close #225.

PR #227, opened 2024-11-29
Make blending (in scaling) respect alpha

abewieland:
When sampling from multiple pixels to implement scaling (only relevant for anti-aliasing), the sampled values should be multiplied by their alpha value to prevent colors "bleeding" from almost-transparent regions. Consider this image, which consists of a square in the center with color #00ff00ff and a border of #ff000001:
alpha_blend

Without multiplying by the alpha, we can get unexpected red in the border region, as seen here with anti-aliasing on (so bicubic):
20241129_14h33m21s_grim
This is because the alpha values are higher there from the green square, but the red component is stronger than it should be.

With the PR gives:
20241129_14h31m30s_grim

### DIFF
--- a/src/pixmap.c
+++ b/src/pixmap.c
@@ -138,10 +138,11 @@ static void scale_average(struct scale_param* sp, size_t start, size_t step)
                 for (src_x = src_x0; src_x <= src_x1; ++src_x) {
                     const argb_t color =
                         sp->src->data[src_y * sp->src->width + src_x];
-                    a += ARGB_GET_A(color);
-                    r += ARGB_GET_R(color);
-                    g += ARGB_GET_G(color);
-                    b += ARGB_GET_B(color);
+                    const uint8_t src_a = ARGB_GET_A(color);
+                    a += src_a;
+                    r += ARGB_GET_R(color) * src_a;
+                    g += ARGB_GET_G(color) * src_a;
+                    b += ARGB_GET_B(color) * src_a;
                     ++count;
                 }
             }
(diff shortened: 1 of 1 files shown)


src/pixmap.c
```diff
@@ -139,9 +139,9 @@ static void scale_average(struct scale_param* sp, size_t start, size_t step)
                     const argb_t color =
                         sp->src->data[src_y * sp->src->width + src_x];
                     a += ARGB_GET_A(color);
-                    r += ARGB_GET_R(color);
-                    g += ARGB_GET_G(color);
-                    b += ARGB_GET_B(color);
+                    r += ARGB_GET_R(color) * ARGB_GET_A(color);
```
artemsen, line 142: I would evaluate `ARGB_GET_A(color)` once to save 0.00000001% =)
I mean
```
const uint8_t src_a = ARGB_GET_A(color);
...
r += ARGB_GET_R(color) * src_a;
```

src/pixmap.c
```diff
@@ -266,17 +274,24 @@ static void scale_bicubic(struct scale_param* sp, size_t start, size_t step)
                 }
             }
 
-            // set pixel
-            for (size_t pc = 0; pc < 4; ++pc) {
+            // set pixel - get alpha first, then rest
+            for (int pc = 3; pc >= 0; --pc) {
```
artemsen, line 278: Please, avoid int/long/short types, use `ssize_t` in this case.

artemsen (commented):
I didn't even think that someone would fix this part of the code.
Thanks!

PR #232, opened 2025-01-03
Small cleanup

abewieland:
Add consistency between the gallery and viewer configs, fix some spelling mistakes, use clamp instead of a mix of min and max for clarity, and ensure all calls to malloc with a sizeof use the variable itself to make future changes to types easier.

### DIFF
--- a/src/gallery.c
+++ b/src/gallery.c
@@ -13,25 +13,25 @@
 #include <stdlib.h>
 
 // Configuration parameters
-#define CFG_SECTION    "gallery"
-#define CFG_SIZE       "size"
-#define CFG_SIZE_DEF   200
-#define CFG_CACHE      "cache"
-#define CFG_CACHE_DEF  100
-#define CFG_FILL       "fill"
-#define CFG_FILL_DEF   true
-#define CFG_WINDOW     "window"
-#define CFG_WINDOW_DEF ARGB(0, 0, 0, 0)
-#define CFG_BACKGR     "background"
-#define CFG_BACKGR_DEF ARGB(0xff, 0x20, 0x20, 0x20)
-#define CFG_SELECT     "select"
-#define CFG_SELECT_DEF ARGB(0xff, 0x40, 0x40, 0x40)
-#define CFG_BORDER     "border"
-#define CFG_BORDER_DEF ARGB(0xff, 0, 0, 0)
-#define CFG_SHADOW     "shadow"
-#define CFG_SHADOW_DEF ARGB(0xff, 0, 0, 0)
-#define CFG_AA         "antialiasing"
-#define CFG_AA_DEF     false
+#define CFG_SECTION          "gallery"
+#define CFG_SIZE             "size"
+#define CFG_SIZE_DEF         200
+#define CFG_CACHE            "cache"
+#define CFG_CACHE_DEF        100
+#define CFG_FILL             "fill"
+#define CFG_FILL_DEF         true
+#define CFG_ANTIALIASING     "antialiasing"
+#define CFG_ANTIALIASING_DEF false
+#define CFG_WINDOW           "window"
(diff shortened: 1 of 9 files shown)


src/gallery.h
```diff
@@ -7,6 +7,18 @@
 #include "config.h"
 #include "event.h"
 
+// Configuration parameters
```
artemsen, line 10: These definitions were deliberately placed in the c-file to reduce their scope.
Now they have become global, but are not used anywhere except in gallery.c.

src/pixmap.h
```diff
@@ -36,8 +36,14 @@ typedef uint32_t argb_t;
 #define ABGR_TO_ARGB(c) \
     ((c & 0xff00ff00) | ARGB_SET_R(ARGB_GET_B(c)) | ARGB_SET_B(ARGB_GET_R(c)))
 
+// Three useful integer functions (all evaluate arguments more than once)
 #define max(a, b) ((a) > (b) ? (a) : (b))
 #define min(a, b) ((a) < (b) ? (a) : (b))
+// On FreeBSD, this conflicts with std::clamp in the tests (which use C++),
+// so just don't define it there
+#ifndef __cplusplus
+#define clamp(a, low, high) (min((high), max((a), (low))))
```
artemsen, line 45: I would move this macro to pixmap.c, it is not used anywhere else anyway.
This will solve the name conflict problem.

artemsen (changes requested):
Thank you!
But please next time split the improvements into different commits to separate typos and functional changes.

src/gallery.h
abewieland: Ah sorry, I didn't see that some of the VIEWER configs were used in `main.c`. I can easily move them back. Alternatively, we could keep both global and use them in `config.c` to reduce the likelihood of strings not matching between them (`config.c` and locations like `viewer.c`, `gallery.c`, etc.) due to typos.

src/pixmap.h
abewieland: Sure thing. I put it there because it seems at the same scope as min and max (and this was just a brief PR of stuff I noticed outside of the scaling while overhauling that). I'll rebase that here quick.

src/gallery.c
```diff
@@ -12,26 +12,28 @@
...
-#define CFG_BACKGR_DEF ARGB(0xff, 0x20, 0x20, 0x20)
-#define CFG_SELECT     "select"
-#define CFG_SELECT_DEF ARGB(0xff, 0x40, 0x40, 0x40)
-#define CFG_BORDER     "border"
-#define CFG_BORDER_DEF ARGB(0xff, 0, 0, 0)
-#define CFG_SHADOW     "shadow"
-#define CFG_SHADOW_DEF ARGB(0xff, 0, 0, 0)
-#define CFG_AA         "antialiasing"
-#define CFG_AA_DEF     false
+// Configuration parameter names
```
artemsen, line 15: Sorry, but I don't like this block of changes.
1. It became less obvious that the name referred to settings. In the case of `CFG_WINDOW`, you can guess that it is related to the configuration. But `GALLERY_WINDOW` is something abstract.
2. The GALLERY prefix is ​​redundant - the scope of these macros is already limited to the gallery source file.
3. The default value of the settings is now stored separately from the name, which makes it difficult to understand. It is easier to read CFG_WINDOW and CFG_WINDOW_DEF side by side than to compare GALLERY_WINDOW and CFG_WINDOW_DEF, especially since they are located in different places.

artemsen (changes requested):

src/gallery.c
abewieland: Yeah, I understand the reasoning. I had just made those changes (for consistency with viewer) long ago and wanted to reduce the difficulty of the rebase (in the last push). I've just pushed again, let me know what you think.

PR #233, opened 2025-01-03
Overhaul scaling

abewieland:
Finally got enough time to work more on antialiasing:

The new version implements 1D convolution kernels, then uses them (first in the horizontal direction, then in the vertical) to implement 2D image scaling. A lot of different optimizations were tried; some made it into this version (including fixed point, storing certain parts of the kernels contiguously, and using truncation rather than `round`). Others which didn't help were excluded.

The new implementation implements the following scaling methods:
1. Nearest neighbor
2. Box (the simplest kernel: averages in a box on downscale, produces nearest neighbor on upscale)
3. Bilinear
4. Bicubic (specifically Catmull-Rom)
5. Magic Kernel Sharp (with the 2013 approximation)

For a direct comparison, these are the same as the following ImageMagick filters:
1. Point
2. Box
3. Triangle
4. CatRom
5. MagicKernelSharp2013

The old version implemented the following:
1. Nearest neighbor
2. Box (it was called average)
3. Bicubic (specifically Catmull-Rom, but only on upscale - nearest neighbor on downscale)

The new implementation has a couple of advantages:
1. Since it uses kernels, this makes implementing new scaling methods (should that become necessary) much easier - almost every other image manipulation program uses kernels, so their code is more helpful and one really just needs to implement a window function and provide the window itself.
2. The implementation is somewhat optimized - it's faster (often at least 2x, sometimes more) than the previous in almost all cases (more on that below), and there are TODOs throughout (some of which I may investigate later) which could perhaps provide even better performance.
3. In some senses, the code is simpler; compared to the old bicubic implementation, for example, I'd claim it's easier to understand which scaling method is in use, for example.
4. It produces more correct scales. As compared with ImageMagick, most color values (for everything but MKS) are within 4 (with MKS they're within 9). This is an error of only 3% (for MKS) or lower, and could be improved by storing more bits between the passes (ImageMagick, like most image manipulators, uses floating point, so has greater accuracy).

Of course, it also has some disadvantages:
1. The current implementation (when using the box method) is somewhat slower than the old average method when downscaling rather small images. If the intermediate were stored multiplied with alpha, I presume this advantage would go away.
2. The code is still somewhat complex, since it's undergone several optimization passes.
3. It uses more memory during scales (once scaling is over, everything is freed), because it needs to store the horizontal and vertical kernels and the intermediate pixmap. It won't use more memory than the output pixmap, though, so it isn't a big increase.

I've left this in draft since I presume it'll still need some changes before being merged, but I thought I'd at least finally put it out there.


artemsen (changes requested):

artemsen (approved):

artemsen:
LGTM, thank you!

abewieland:
The last force push was just to squash the commits into one with the proper sign-off.

artemsen:
Oops... Looks like we broke the nearest mode.
Images that are larger than the window size now extend beyond the borders.

abewieland:
Do you have an example? I'm not entirely sure what you mean by extending beyond the borders. For example, I tried the following:
```
wget -O- https://upload.wikimedia.org/wikipedia/commons/4/47/PNG_transparency_demonstration_1.png | ./build/swayimg -p 500,300 -w 300,300 -c viewer.window=#000000ff -c info.show=no -
```

That gives this on startup:
20250105_10h02m35s_grim

And after zooming in quite a bit:
20250105_10h02m40s_grim

Both windows are 300x300 as expected (I measured in GIMP), and the image is larger (662x497).

artemsen:
These images are displayed at 99% scale.

AA ON:
aa_on

AA OFF:
aa_off

Original image:
original

abewieland:
I see, probably rounding errors with the fixed-point representation; let me look into it a little more.

PR #234, opened 2025-01-05
Avoid drift in nearest-neighbor

abewieland:
Previously, pictures were sometimes cut off due to insufficient precision in the fixed-point (especially when the scale was very nearly 1), or had some jitter when turning anti-aliasing on and off (especially when downscaling). This helps with both those issues, though cannot entirely eliminate them (without introducing significant slowdowns with floor and ceil, min and max in the loop).


artemsen:
Works as expected!
Thank you!

PR #240, opened 2025-01-20
add keep_position, anchoring based on config position

uwidev:
This is mainly to extend the usefulness of `keep_zoom`. 

When comparing images with `keep_zoom` enabled, users might adjust the position and zoom. When they change to the comparing image, the position resets to whatever `position` was set to. This doesn't allow for good A/B comparison workflows at the moment, as the only way to do efficient comparisons is to never adjust position. This will allow position to be preserved.

### DIFF
--- a/src/viewer.c
+++ b/src/viewer.c
@@ -77,6 +77,7 @@ static const char* position_names[] = {
 
 /** Viewer context. */
 struct viewer {
+    bool on_first_image;         ///< Differentiate first and others
     ssize_t img_x, img_y;        ///< Top left corner of the image
     size_t frame;                ///< Index of the current frame
     argb_t image_bkg;            ///< Image background mode/color
@@ -88,6 +89,9 @@ struct viewer {
     bool keep_zoom;              ///< Keep absolute zoom across images
     enum position position;      ///< Initial position
     float scale;                 ///< Current scale factor of the image
+
+    size_t prev_image_width;  ///< Previously loaded image's width
+    size_t prev_image_height; ///< Previously loaded image's height
 
     bool animation_enable; ///< Animation enable/disable
     int animation_fd;      ///< Animation timer
@@ -325,6 +329,8 @@ static void zoom_image(const char* params)
         const float step = (ctx.scale / 100) * percent;
         const double center_x = wnd_half_w / ctx.scale - ctx.img_x / ctx.scale;
         const double center_y = wnd_half_h / ctx.scale - ctx.img_y / ctx.scale;
+        const struct image* img = fetcher_current();
+        const struct pixmap* pm = &img->frames[ctx.frame].pm;
 
         if (percent > 0) {
             ctx.scale += step;
@@ -332,8 +338,6 @@ static void zoom_image(const char* params)
                 ctx.scale = MAX_SCALE;
             }
         } else {
-            const struct image* img = fetcher_current();
-            const struct pixmap* pm = &img->frames[ctx.frame].pm;
             const float scale_w = (float)MIN_SCALE / pm->width;
             const float scale_h = (float)MIN_SCALE / pm->height;
             const float scale_min = max(scale_w, scale_h);
@@ -428,12 +432,16 @@ static void reset_state(void)
 {
     const struct image* img = fetcher_current();
     const size_t total_img = image_list_size();
+    const struct pixmap* pm = &img->frames[ctx.frame].pm;
+    const bool eq_dim_prev_img = (pm->width == ctx.prev_image_width &&
+                                  pm->height == ctx.prev_image_height);
 
     ctx.frame = 0;
-    if (!ctx.keep_zoom || ctx.scale == 0) {
+    if (!ctx.keep_zoom || ctx.scale == 0 || !eq_dim_prev_img) {
         scale_image(ctx.scale_init);
+    } else {
+        fixup_position(false);
     }
-    set_position();
 
     ui_set_title(img->name);
     animation_ctl(true);
@@ -474,6 +482,12 @@ static bool skip_image(void)
 static bool next_image(enum action_type direction)
 {
     size_t index = fetcher_current()->index;
+
+    ctx.on_first_image = false;
+
+    const struct pixmap* img = &fetcher_current()->frames[ctx.frame].pm;
+    ctx.prev_image_width = img->width;
+    ctx.prev_image_height = img->height;
 
     do {
         switch (direction) {
@@ -675,6 +689,7 @@ static void apply_action(const struct action* action)
             break;
         case action_mode:
             app_switch_mode(fetcher_current()->index);
+            ctx.on_first_image = true;
             break;
         case action_step_left:
             move_image(true, true, action->params);
@@ -792,6 +807,11 @@ void viewer_init(const struct config* cfg, struct image* image)
     if (ctx.slideshow_fd != -1) {
         app_watch(ctx.slideshow_fd, on_slideshow_timer, NULL);
     }
+
+    // hacks
+    ctx.on_first_image = true;
+    ctx.prev_image_width = 0;
+    ctx.prev_image_height = 0;
 
     fetcher_init(image, history, preload);
 }


uwidev:
I'm having some issues with it crashing, will investigate and update further.

uwidev:
Images are anchored based on the top left `(0,0)`. I was considering having that anchor be based on the config's `position`. I could see if I can implement anchoring...

uwidev:
`keep_position` should now be anchored to whatever `position` is set to. 

Scaling is also anchored to position to make it consistent.

New position setting `absolutecenter` to base position to anchor to window center (original functionality). `center` to anchor to image center.

uwidev:
There's an issue where when `keep_position = yes` and `fixed = no`, the first image that is loaded is aligned to the top left of the screen. What needs to happen is that for the very first image that loads when entering the viewer, the image is scaled according to `scale_init`, aka the config `scale`.

I'm pretty brain dead right now. Any help on where to look would be appreciated.

uwidev:
Patched first image not centering.

There is now a `after_first_image_loaded`, set to `false` during viewer's init function, and set to `true` towards the end of loading the next (first) image, but before calling `reset_state()`. I think I was having issues putting it after resetting the state, so I just put it before and it just works. Good enough for now, until next time.

extra/swayimgrc.5
```diff
@@ -102,11 +102,15 @@ Default image scale, valid modes are:
...
+.IP "\fBkeep_position\fR\fR = \fI[yes|no]\fR"
+Keep absolute position across images, \fIno\fR by default.
+.\" ----------------------------------------------------------------------------
 .IP "\fBposition\fR = \fIPOSITION\fR"
 Initial image position, valid modes are:
 .nf
 \fItop\fR: move image to top and center by width;
-\fIcenter\fR: center by width and height (default);
+\fIcenter\fR: center by image width and height;
+\fIabsolutecenter\fR: center by window width and height (default);
```
artemsen, line 113: This looks confusing. What the difference between center and absolute center? Is the first center relative now?

src/viewer.c
```diff
@@ -160,46 +171,91 @@ static void set_position(void)
     const size_t wnd_width = ui_get_width();
     const size_t wnd_height = ui_get_height();
 
-    switch (ctx.position) {
-        case position_top:
-            ctx.img_y = 0;
-            ctx.img_x = wnd_width / 2 - (ctx.scale * pm->width) / 2;
-            break;
-        case position_center:
+    if (ctx.keep_position){
```
artemsen, line 174: Please, use clang-format.

src/viewer.c
```diff
@@ -160,46 +171,91 @@ static void set_position(void)
...
-            break;
-        case position_bottom_right:
-            ctx.img_y = wnd_height - ctx.scale * pm->height;
-            ctx.img_x = wnd_width - ctx.scale * pm->width;
-            break;
+        }
+
+        switch (ctx.position) {
+            case position_top:
+                ctx.img_x += ctx.scale * ((int)ctx.last_width - (int)pm->width) / 2;
```
artemsen, line 182: Please use ssize_t instead of int here.

src/viewer.c
```diff
@@ -77,6 +79,7 @@ static const char* position_names[] = {
 
 /** Viewer context. */
 struct viewer {
+    bool after_first_image_load; ///< Differentiate first and others
```
artemsen, line 82: Looks like a hack =)

artemsen (changes requested):
Could you describe the use case for this?
What if images have different sizes?

artemsen:
I think we could combine `keep_zoom` and `keep_position` into one option.
And may be this combined option can replace `absolutecenter` (which I don't like, to be honest =)).

extra/swayimgrc.5
uwidev: The way I saw it is that positions are based on the image. However, zooming was based on the center of the window rather than the center of the image. Since this anchoring thing was based on the image instead of the window, I thought it would be a good idea to change zooming to be based on the image center rather than the window center.

I can rename absolutecetner->windowcenter, that might make it more clear; or I can have center->imagecenter and revert the other center. Whatever you think is more clear.

uwidev: Removed absolute center. Zooming anchored to various parts of the image feels unintuitive and I feel silly for implementing it.

uwidev:
Merged `keep_position` into `keep_zoom`. Adjusted behavior such that it only keeps zoom if the image switched to is the same dimension (it can be disorienting if it's a completely different image of different dimensions). This feels like the most comfortable mode of operation.

Though some people might not want it like this, and to always keep zoom regardless of if it's a completely different image? What are your thoughts on this?

src/ui.c
```diff
@@ -440,11 +440,14 @@ static void handle_xdg_toplevel_configure(void* data, struct xdg_toplevel* lvl,
             wp_viewport_set_destination(ctx.wp.viewport, width, height);
         }
         ctx.xdg.initialized = true;
-        app_on_resize();
     }
 
-    if (reset_buffers && !recreate_buffers()) {
-        app_exit(1);
+    if (reset_buffers) {
```
artemsen, line 445: You should rebase instead of copy/paste master's commit.
And please squash your commits.

src/viewer.c
```diff
@@ -578,8 +642,8 @@ static void draw_image(struct pixmap* wnd)
     // clear image background
     if (img->alpha) {
         if (ctx.image_bkg == GRID_BKGID) {
-            pixmap_grid(wnd, ctx.img_x, ctx.img_y, width, height,
-                        GRID_STEP, GRID_COLOR1, GRID_COLOR2);
+            pixmap_grid(wnd, ctx.img_x, ctx.img_y, width, height, GRID_STEP,
```
artemsen, line 645: This change is not needed.

src/viewer.c
```diff
@@ -635,6 +699,7 @@ static void on_resize(void)
 {
     fixup_position(false);
     reset_state();
+    // ctx.after_first_image_load = true;
```
artemsen, line 702: Commented source code.

artemsen (changes requested):

src/viewer.c
uwidev: clang-format was moving this down. Fixed it manually

src/ui.c
uwidev: I thought I rebased it but seems like it did something funky. Squashed.

artemsen:
I dug a little deeper into the problem and found a simpler way to solve this problem.
In addition, this solution allows for more correct processing of images that differ in size - scaling is not reset.
@uwidev, please look at [#242](https://github.com/artemsen/swayimg/pull/242)

uwidev:
I made those adjustments on my end. I removed the `set_position()` you added, because `scale_image()` already calls it at the end, so it was redundant. I also set `fixup_position(true)` to `false` because during use, it's disorienting to have images moved (when `fixed=no`).

I was thinking more about A/B image testing and I ran into a few problems.

Case 1: Images A/B are visually similar but dimensional the same. Zoom is retained
Case 2: Images A/B are exactly the same content-wise, but cropped differently, and this of differing dimensions. Zoom is reset.

In this current iteration, zoom will reset when images are not dimensionally equal. But for case 2, it feels *off*. 

The point of A/B testing is to, of course, compare the images. Specifically, we are comparing the overall shape of the image, its colors, fine details, etc. The first case assumes the images have mostly the same positional data. However, when images are cropped differently, or taken from a slightly different angle, the positional data can change drastically, which can be disorienting when the zoom is kept. In cases like this, perhaps it would be better to keep zoom absolute? It feels like having this keep zoom be dynamic to image size isn't so good for Case 2, but is good for Case 1. What are your thoughts on this?

An more complex solution to truly fixing position is to align it based on best matching/overlapping image details. I think [Hydrus](https://github.com/hydrusnetwork/hydrus/tree/master) does this for images marked as potential [duplicates](https://hydrusnetwork.github.io/hydrus/duplicates.html) (maybe different angle/cropped different). If memory serves me right, it aligns the most matched/defining parts of the images for A/B testing . Implementing this might be beyond the scope of this project, but it's the most interesting implementation of "keep position" as instead of relying on an anchor on the image, it anchors the image based on the most prominently similar details.

artemsen:
@uwidev, thanks for your improvements, but I think that to solve the current problem it is enough to use the simpler solution described in #242. I added you as a co-author in the commit, but if you are against it - let me know.
Handling cases with comparison of images of different sizes is a non-trivial task related to recognition of their content, which is beyond the scope of an image viewer. Therefore, I believe that further research into this issue is not worthwhile.

uwidev:
I understand that swayimg is meant to be strictly an image viewer, but having the ability to immediately switch from one image to the next at the very least brings the emergent feature of comparing images, even such capabilities isn't further developed. For some, a simple flip-flop is enough. It's just that since keeping zoom is now not dynamic to the image resolution, it no longer feels like an intuitive default mode of use, as I find myself constantly doing a zoom fit.

Would you at least consider having a bind to toggle keep zoom so users can manually lock the zoom? Having the bind would at least let it be a temporary mode to switch to as needed.

artemsen:
Cropped images are now handled correctly.
The zoom keeping mode is switched by Alt-z.

Issue #248, opened 2025-02-07
All images are displayed with <100% opacity


litoj:
Hello, I've just noticed that I can see through the photo displayed with swayimg in viewer mode (gallery works fine). This is despite the photo being a JPG (no transparency possible) and config set to `viewer.window=#000000`. The result looks like this:

![Image](https://github.com/user-attachments/assets/73ca6b3a-019c-4c7f-8bc4-ee04b7657a5f)

In the horizontal middle (most obvious at the top) you can see the gap between my windows behind swayimg. However, it is not visible through the border bars (the `viewer.window` works as expected).

Trying `viewer.transparency=#000000` yielded no different results.

artemsen:
It looks like the transparency is set at the system level.
Sorry, I can't reproduce it.

litoj:
Yes, I have sway set to have transparent background. But shouldn't the images be shown with 100% opacity, so the background wouldn't be visible anyway?

litoj:
Antialiasing set to none fixes the issue. While antialiasing set to `bilinear` makes it very obvious. Using that, I found that the transparency happens for images at any scale `!=1`, but is less noticeable when using other antialiasing methods.

For example with the `box` antialiasing method, I noticed strange transparent lines forming a grid appearing over the image: (most noticeable in the right bottom darker areas)

![Image](https://github.com/user-attachments/assets/765539ae-7491-4297-8674-ad8c79ba83d3)

artemsen:
> Yes, I have sway set to have transparent background. But I'd expect the images to be shown with 100% opacity, so the background wouldn't be visible anyway, right?

I don't think so, because Image opacity != window opacity.
If you set the opacity of the swayimg window, it will have a transparent background even if the image is not transparent at all.

> Antialiasing set to on in version 3.6 and earlier doesn't exhibit this issue. It started in version 3.7 with the addition of different antialiasing methods.

There were some inaccuracies with alpha blending that have been fixed in 3.7.

litoj:
Since I provided more detailed steps to reproduce the issue, is it still unconfirmed?

artemsen:
Yes, I tried it. I set system-wide transparency and got the same effect.
But I think it works as expected. If you want transparent windows, you will get transparent windows. It does not matter what settings you use in swayimg, because this transparency comes from the system.
Or may be I don't understand something, sorry =)

litoj:
[...]

PR #250, opened 2025-02-15
Ensure kernel weights sum to 1 in pixmap_scale

abewieland:
Moved away from truncation and added fixed-point sum to ensure the weights all sum to 1 (that is, the fixed-point multiplier), avoiding situations where opaque images gain some transparency.

Related to #248.


artemsen:
Thank you!

PR #258, opened 2025-03-05
Offset window position and size according to sway border thickness

Toraalv:
First of all, this is my first PR ever, I apologise for any errors.

So the problem I encountered was that when I open an image, the application window is wrongly offset by the thickness of my sway borders.

Relevant configuration to reproduce the problem:

- Enable sway borders, together with smart borders
- Enable swayimg window decorations
- Open any image.

Image of the problem (I have set sway borders to 32px thick to easier display the problem):
bild
bild

In a non-floating window:
bild

My PR fixes this. I have tested all four combinations of smart border on or off and swayimg's window decoration on or off.

I am by far not a C programmer, but I have done my best to respect your source code. I have also written some comments in the code that are only for better explaining this PR, feel free to remove them.

### DIFF
--- a/src/application.c
+++ b/src/application.c
@@ -76,7 +76,7 @@ static struct application ctx;
 /**
  * Setup window position via Sway IPC.
  */
-static void sway_setup(void)
+static void sway_setup(const struct config* cfg)
 {
     struct wndrect parent;
     bool fullscreen;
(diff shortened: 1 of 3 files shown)


artemsen:
Thank you for the PR!
I understood the problem, but the proposed solution has problems with the architecture: sway.h should not work with the config. Let me fix it myself.

PS. Nice photo. I'm a big Ozzy fan =)

artemsen:
@Toraalv, could you please check the latest master?

Toraalv:
It looks good to me, but I have one question. In a floating window, swayimg now appears inside the window, not on top of it. See image:
bild
(dark grey is parent window, light grey is swayimg window)

Is this intentional behavior? Either works of course, I just assumed it was supposed to go on top of the parent window.

artemsen:
Yes, I think that's correct. Both windows have frames, one of them is placed in the dimensions of the other. This can be avoided by specifying the border size to 0 for the window of class swayimg.

Toraalv:
All right! Thank you very much.

PR #264, opened 2025-03-16
Added sort order by modified time

gcorrad:
Hello!, excellent project. I hope this contribution is useful, I added the modification date as a sorting option and the reverse option works as a flag for both alpha and mtime sorting.

### DIFF
--- a/src/imagelist.c
+++ b/src/imagelist.c
@@ -17,12 +17,20 @@
 #include <time.h>
 #include <unistd.h>
 
+/** Image list array entry. */
+typedef struct _image_src {
+    char *source;          ///< Entry name
+    time_t time;           ///< File time
+    time_t dirtime;        ///< Directory time
+} t_image_src;
+
 /** Context of the image list (which is actually an array). */
 struct image_list {
-    char** sources;        ///< Array of entries
+    t_image_src *sources;  ///< Array of entries
     size_t capacity;       ///< Number of allocated entries (size of array)
     size_t size;           ///< Number of entries in array
     enum list_order order; ///< File list order
+    bool reverse;          ///< Reverse order flag
     bool loop;             ///< File list loop mode
     bool recursive;        ///< Read directories recursively
     bool all_files;        ///< Open all files from the same directory
@@ -35,7 +43,7 @@ static struct image_list ctx;
 static const char* order_names[] = {
     [order_none] = "none",
     [order_alpha] = "alpha",
-    [order_reverse] = "reverse",
+    [order_time] = "time",
     [order_random] = "random",
 };
 
@@ -131,11 +139,11 @@ static size_t absolute_path(const char* source, char* path, size_t path_max)
  * Add new entry to the list.
  * @param source image data source to add
  */
-static void add_entry(const char* source)
+static void add_entry(const char* source, struct stat *st, struct stat *dir_st)
 {
     // check for duplicates
     for (size_t i = 0; i < ctx.size; ++i) {
-        if (strcmp(ctx.sources[i], source) == 0) {
+        if (strcmp(ctx.sources[i].source, source) == 0) {
             return;
         }
     }
@@ -143,7 +151,7 @@ static void add_entry(const char* source)
     // relocate array, if needed
     if (ctx.size + 1 >= ctx.capacity) {
         const size_t cap = ctx.capacity ? ctx.capacity * 2 : 4;
-        char** ptr = realloc(ctx.sources, cap * sizeof(*ctx.sources));
+        t_image_src *ptr = realloc(ctx.sources, cap * sizeof(t_image_src));
         if (!ptr) {
             return;
         }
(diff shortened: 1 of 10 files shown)


src/imagelist.c
```diff
@@ -299,12 +344,14 @@ void image_list_destroy(void)
 
 void image_list_add(const char* source)
 {
-    struct stat st;
+    struct stat st, dir_st;
+    memset(&st,     0, sizeof(struct stat));
```
artemsen, line 348: Please, use clang-format to format the code. The root dir contains the file with rules for that.

src/main.c
```diff
@@ -28,6 +28,7 @@ static const struct cmdarg arguments[] = {
     { 'g', "gallery",    NULL,    "start in gallery mode" },
     { 'r', "recursive",  NULL,    "read directories recursively" },
     { 'o', "order",      "ORDER", "set sort order for image list: none/[alpha]/reverse/random" },
+    { 'R', "reverse",    NULL,    "reverse sort order" },
```
artemsen, line 31: This change requires modification of man and bash/zsh completion scripts.
It's not used very often and can be set via `-c`, so it can be omitted.

src/imagelist.c
```diff
@@ -289,7 +334,7 @@ void image_list_init(const struct config* cfg)
 void image_list_destroy(void)
 {
     for (size_t i = 0; i < ctx.size; ++i) {
-        free(ctx.sources[i]);
+        free(ctx.sources[i].source);
```
artemsen, line 337: It works because source addres is equal to image_src struct. Looks strange that you free string but not the whole struct.

artemsen (changes requested):

src/imagelist.c
gcorrad: in add_entry function alloc memory for file path
    // add new entry
    ctx.sources[ctx.size].source = str_dup(source, NULL);

..but before do that allocate memory for the array if needed
    // relocate array, if needed
    if (ctx.size + 1 >= ctx.capacity) {
        const size_t cap = ctx.capacity ? ctx.capacity * 2 : 4;
        struct image_src *ptr = realloc(ctx.sources, cap * sizeof(struct image_src));

and when release memory release file paths allocated and then the array
    for (size_t i = 0; i < ctx.size; ++i) {
        free(ctx.sources[i].source);
    }
    free(ctx.sources);

i think is ok..

artemsen (changes requested):

src/imagelist.c
```diff
@@ -152,8 +164,24 @@ static void add_entry(const char* source)
     }
 
     // add new entry
-    ctx.sources[ctx.size] = str_dup(source, NULL);
-    if (ctx.sources[ctx.size]) {
+    ctx.sources[ctx.size].source = str_dup(source, NULL);
+    if (ctx.sources[ctx.size].source) {
+        switch(ctx.order) {
+        case order_ctime:
+            ctx.sources[ctx.size].data.time = st->st_ctime;
```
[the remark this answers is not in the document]
gcorrad: i will

src/imagelist.c
```diff
@@ -17,12 +17,22 @@
 #include <time.h>
 #include <unistd.h>
 
+/** Image list array entry. */
+struct image_src {
+    char *source;          ///< Entry name
+    union {
+        time_t time;
+        size_t size;
+    } data;
```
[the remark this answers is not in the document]
gcorrad: i will

src/imagelist.c
```diff
@@ -265,22 +293,45 @@ static size_t next_dir(size_t start, bool forward)
...
+/**
+ * Compare sources callback for `qsort`.
+ * @return negative if a < b, positive if a > b, 0 otherwise
+ */
+static int compare_size(const void* a, const void* b)
+{
+    size_t sa = ((struct image_src*)a)->data.size;
+    size_t sb = ((struct image_src*)b)->data.size;
+
+    if (sa < sb) return ctx.reverse?  1: -1;
```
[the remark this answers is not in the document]
gcorrad: please can you paste here the correct format for these lines? i don't know so much clang-format

src/imagelist.c
```diff
@@ -265,22 +288,53 @@ static size_t next_dir(size_t start, bool forward)
...
 /**
  * Compare sources callback for `qsort`.
- * @return negative if a > b, positive if a < b, 0 otherwise
+ * @return negative if a < b, positive if a > b, 0 otherwise
  */
-static int compare_reverse(const void* a, const void* b)
+static int compare_time(const void* a, const void* b)
 {
-    return -strcoll(*(const char**)a, *(const char**)b);
+    time_t ta = ((struct image_src*)a)->time;
```
artemsen, line 303: I would make it simpler, something like
```c
const time_t ta = ((struct image_src*)a)->time;
const time_t tb = ((struct image_src*)b)->time;
return ctx.reverse ? ta - tb : tb - ta;
```

src/main.c
```diff
@@ -27,7 +27,7 @@ struct cmdarg {
 static const struct cmdarg arguments[] = {
     { 'g', "gallery",    NULL,    "start in gallery mode" },
     { 'r', "recursive",  NULL,    "read directories recursively" },
-    { 'o', "order",      "ORDER", "set sort order for image list: none/[alpha]/reverse/random" },
+    { 'o', "order",      "ORDER", "set sort order for image list: none/alpha/mtime/size/random" },
```
artemsen, line 30: You shouldn't have removed the square brackets, they show the default value.

src/imagelist.c
```diff
@@ -265,22 +288,53 @@ static size_t next_dir(size_t start, bool forward)
  */
 static int compare_alpha(const void* a, const void* b)
 {
-    return strcoll(*(const char**)a, *(const char**)b);
+    int cmp =
```
artemsen, line 291: Can be `const`

artemsen (changes requested):

PR #268, opened 2025-03-30
Add numeric sort order

duxovni:
The current `alpha` ordering sorts filenames with numbers out-of-order when they don't have leading zeroes. I've added a new order option that uses `strverscmp()` to produce a more natural numeric ordering.

### DIFF
--- a/extra/swayimg.1
+++ b/extra/swayimg.1
@@ -42,6 +42,7 @@ Set order of the image list:
 .nf
 \fInone\fR: unsorted, order is system depended;
 \fIalpha\fR: sorted alphabetically (default);
+\fInumeric\fR: sorted numerically;
 \fImtime\fR: sorted by file modification time;
 \fIsize\fR: sorted by file size;
 \fIrandom\fR: randomize list.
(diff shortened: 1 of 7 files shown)


duxovni:
(Not sure why the FreeBSD check is failing; https://man.freebsd.org/cgi/man.cgi?query=strverscmp says `strverscmp` was added to `string.h` starting in FreeBSD 13.2.)

artemsen:
`strverscmp` is a GNU extension to libc.
That's why you have to define _GNU_SOURCE.
That's why there is a problem with build in FreeBSD.
That's why we may have compatibility issues in other systems.

Sorry, I can't accept this PR because it binds the project to a specific libc implementation.

duxovni:
Understood, no worries. I don't have the spare time to implement a numeric comparison function from scratch, but hopefully someone will be able to eventually; I still think it's a valuable feature to have.

artemsen:
> I still think it's a valuable feature to have.

I agree.
I created the [issue](https://github.com/artemsen/swayimg/issues/270) so that I wouldn't forget about it.

As a workaround you can use `sort` (but it work only with disabled the [all](https://github.com/artemsen/swayimg/blob/10e46b150cdbd2195fcbe1d74188c7e2ae152093/extra/swayimgrc#L99) mode)
```
ls | sort --numeric-sort | xargs swayimg -o none
```

Issue #266, opened 2025-03-28
font size doesn't scale to the wayland output scale


lilydjwg:
E.g. when I set font size to 24, it looks good for me on my 2x scale monitor, but with the same config, on my 1x laptop monitor the text looks large. (Both of them have character height somewhat near 32 physical pixels.)

artemsen:
Support for the wayland fractional scaling was added in swayimg v.3.7.
Which version of swayimg do you use? And what WM?

lilydjwg:
Yes, fractional scaling is supported and the image itself is sharp. However the font rendering part isn't aware of it. It always uses 96dpi without any scaling:

https://github.com/artemsen/swayimg/blob/10e46b150cdbd2195fcbe1d74188c7e2ae152093/src/font.c#L140

I'm using swayimg 3.9 and Wayfire.

artemsen:
It doesn't matter what dpi is used in fonts, because we have [scaled buffer](https://github.com/artemsen/swayimg/blob/10e46b150cdbd2195fcbe1d74188c7e2ae152093/src/ui.c#L849) which is greater than the window.
Sorry but I can't reproduce the bug, both my displays (1x and 2x) in my environment (Sway WM) show the same font size.

I'll try to reproduce this in a VM with Wayfire when I have time.

lilydjwg:
> both my displays (1x and 2x) in my environment (Sway WM) show the same font size.

I think this is the problem. The font size is the same with reference to the image (of the same scale) (or the same monitor). The text always occupies the same physical pixels no matter what the scale is, which means it has different sizes on monitors with different pixel density; i.e. with larger pixels, the text is larger.

I expect text to be too small if I set my monitor to 1x scale when 2x should be used.

I've tested with Sway & KDE in a VM; they are the same.

artemsen:
Hmm... Then it works as designed. I expect that physical font size will be the same on different displays with different scale factors.

lilydjwg:
It depends on how you measure the size. In pixels yes they are the same. But in centimeters, no, either one of my monitor has too small text, or the other one has too large text.

artemsen:
Yes, by physical size I meant centimeters. If I have 2x on one display and 1x on the other, the font should have the same size in centimeters on both displays. And in my environment this is indeed the case, but didn't set any DPI settings, just scale factors.

lilydjwg:
Then something seems wrong.

[...]

PR #272, opened 2025-04-02
scale text using wayland scale factor

lilydjwg:
This patch aims to solve #266 to scale text using wayland scale factor like all other wayland applications do.

I can add missing doc comments if this is going to be accepted. Also, please tell me if any refactor is desired.


lilydjwg:
No, it should have fixed them instead of breaking them. Did you capture the screenshots pixel-to-pixel with 2x scale and 1x scale respectively? If so, it is expected, because 2x displays have denser pixels, the more-pixel text will look the size as the 1x displays with less-pixel ones.

You need actual lodpi & hidpi displays to see the result. Changing scales alone won't look right.

artemsen:
> No, it should have fixed them instead of breaking them.

Did you try to test it yourself?

> Did you capture the screenshots pixel-to-pixel with 2x scale and 1x scale respectively?

These screenshot taken from the same display but with different scale factors. The master version of swayimg shows fonts of the same size regardless of the scale factor.

artemsen:
> Did you notice that other applications (e.g. Firefox) have larger text with 2x scale?

Yes, of course. I also expect Firefox running on different displays to have the same font size regardless of DPI, because that's what the scaling factor is for. And I expect the same from swayimg: no matter what display it's running on, I want the same font size (like Firefox).

artemsen:
Sorry, I looked at the code more carefully and maybe you're right =)

src/font.c
```diff
@@ -21,6 +21,7 @@
 struct font {
     FT_Library lib;    ///< Font lib instance
     FT_Face face;      ///< Font face instance
+    size_t font_size;
```
artemsen, line 24: Just `size`

src/font.c
```diff
@@ -137,6 +138,7 @@ void font_init(const struct config* cfg)
 
     // set font size
     font_size = config_get_num(cfg, CFG_FONT, CFG_FONT_SIZE, 1, 256);
```
artemsen, line 140: `ctx.size = config_get_num(...`

src/font.h
```diff
@@ -19,6 +19,8 @@ struct text_surface {
  */
 void font_init(const struct config* cfg);
 
+void font_on_scale(double scale);
```
artemsen, line 22: `font_set_scale`

src/info.c
```diff
@@ -454,6 +454,11 @@ void info_init(const struct config* cfg)
         config_get_num(cfg, CFG_INFO, CFG_INFO_STIMEOUT, 0, 1024);
     timeout_init(&ctx.status);
 
+    info_on_scale();
+}
+
+void info_on_scale(void)
```
artemsen, line 460: The key size has been updated, but all values ​​remain the same.

src/ui.c
```diff
@@ -8,6 +8,8 @@
 #include "array.h"
 #include "buildcfg.h"
 #include "wndbuf.h"
+#include "font.h"
```
artemsen, line 11: Please use clang-format, it will arrange them in the right order.

artemsen (changes requested):

src/info.c
lilydjwg: Do you mean the values that come after the keys? They'll be redrawn on the following `app_redraw()` call.

PR #275, opened 2025-04-08
Small scaling optimization

abewieland:
Avoid multiplying (and importantly, dividing) by the alpha channel if we already know the image is opaque.

For images with no alpha channel (like jpgs, which already tend to be larger), the change resulted in a 10-50% speed improvement in my testing, depending on the anti-aliasing mode and scale factor.

Notably, however, this means that if an image _does_ have an alpha channel, passing `alpha = false` to `pixmap_scale` will no longer give correct results.


artemsen:
This code already seems like magic to me =)
Thanks!

PR #279, opened 2025-05-04
feat: rerender SVGs at current scale

GenericGuy:
This PR adds a new keybinding (`Alt+R`), which rerenders the currently displayed SVG at the current scale level:

**Before (blurry zoomed-in view):**

2025-05-05-135625_hyprshot

**After (rendered at current scale):**

2025-05-05-135913_hyprshot

I essentially just added the keybinding and created two functions to set and reset the scaling factor for the SVG module. When `Alt+R` is pressed in the viewer, the new scaling factor is applied and the image is reloaded (similar as is done in `reload_current()`, but no position change).

The scaling factor is reset when a new file is loaded. A more elegant solution would be to save this on a per-image basis, but that would introduce image-type-specific variables to the context. I chose the current approach for simplicity and to avoid architectural changes.

I’m still learning the ropes with PRs, so please let me know if there’s anything I can improve.
=> 1. lesson learned: enable pipelines on the fork before creating a PR

### DIFF
--- a/extra/swayimgrc
+++ b/extra/swayimgrc
@@ -183,6 +183,7 @@ Shift+m = flip_horizontal
 a = antialiasing next
 Shift+a = antialiasing prev
 r = reload
+Alt+r = svg_rerender
 i = info
 Shift+Delete = exec rm -f '%' && echo "File removed: %"; skip_file
 Escape = exit
--- a/extra/swayimgrc.5
+++ b/extra/swayimgrc.5
@@ -354,6 +354,7 @@ Predefined names for mouse scroll:
 .IP "\fBrotate_right\fR: rotate image clockwise;"
 .IP "\fBflip_vertical\fR: flip image vertically;"
 .IP "\fBflip_horizontal\fR: flip image horizontally;"
+.IP "\fBsvg_rerender\fR: rerender SVG with current zoom level;"
 .IP "\fBreload\fR: reset cache and reload current image;"
 .IP "\fBantialiasing\fR \fI[MODE]\fR: set anti-aliasing mode or cycle through them (\fInext\fR/\fIprev\fR or mode name);"
 .IP "\fBinfo\fR \fI[MODE]\fR: set info mode or cycle through them (\fIoff\fR/\fIviewer\fR/\fIgallery\fR);"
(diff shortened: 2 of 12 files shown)


artemsen:
> When Alt+R is pressed in the viewer, the new scaling factor is applied and the image is reloaded.

I also thought about rendering scaled svg, but never solved this problem.
This PR is an interesting implementation of such feature. But honestly, I think using an action for this purpose looks like a dirty hack. Someone will definitely create an issue to automatically call the action for all SVG images.

> A more elegant solution would be to save this on a per-image basis, but that would introduce image-type-specific variables to the context. I chose the current approach for simplicity and to avoid architectural changes.

I would prefer architectural changes over workarounds. The proposed implementation has a multithreading issue: loading svg in one thread will break the action in another.

I think we should find a way to rescale SVG images automatically, without user interaction. Any idea how to do this?

GenericGuy:
Yeah, that was my initial goal as well. I've updated the PR with a draft of a potential way to do this.
I'm not entirely sure if this introduces concurrency issues. I'm having a hard time wrapping my head around when the preloader / main loader runs and if the current image is guaranteed to be fully loaded when `draw_image()` is called.
If not, I can only see this work if all accesses of the raw image data are wrapped in mutexes.

Also, this implementation has the same downside most other image viewers have when opening SVGs - if you open a complex SVG the whole viewer is slowed down (initial load, zooming, panning).

artemsen:
I finally found the right way to implement this feature.

> I’m still learning the ropes with PRs

This PR is not how requests are usually reviewed, sorry.
Considering the non-trivial nature of the feature, it is faster for me to implement it myself than to describe the new architecture in human-readable form =)

GenericGuy:
Sorry for not updating the PR, got caught up in some issues that are also present in your solution. Namely that most other functions that alter how the image is displayed are not available with a custom renderer (e.g. rotate, mirror).
I quickly ported my wip to your changes (see my `fix_decoder` branch). It's not perfect yet, but the basic functionality works.
If you think my approach is fine I can try to polish it. As far as I can see the only remaining issues are position shifts when zoomed in and rotating afterwards and misaligned background grid after rotating.

artemsen:
> If you think my approach is fine I can try to polish it.

Look good, create new PR.

> As far as I can see the only remaining issues are position shifts when zoomed in and rotating afterwards and misaligned background grid after rotating.

You should swap width and height in the first image frame.

PR #293, opened 2025-06-02
Remove mention of `-R/--reverse` from man

odrling:
This option is not handled by the CLI.
It was a leftover from #264 where the implementation was found to be [confusing](https://github.com/artemsen/swayimg/pull/264#discussion_r2000410307).


artemsen:
Thank you!

PR #306, opened 2025-06-19
Add antialiasing actions: on, off, toggle

voidbert:
Now, the `antialiasing` config option sets the preferred anti-aliasing type, which can be changed with `antialiasing prev` and `antialiasing next`. The actions `antialiasing on`, `antialiasing off`, and `antialiasing toggle` have also been added, and switch between the no anti-aliasing and the preferred anti-aliasing mode.


voidbert:
@artemsen, before I solve the merge conflict and change the manual, could you please check if this API is OK? I made it so nothing from the previous versions is broken.

artemsen:
You just made things more complicated =)
I don't think we need all these switches at all. These hacks come from the times when AA worked extremely slowly, now it is no longer a problem.
It is unlikely that anyone needs to run a program with AA turned off and then turn it on manually, so I would just remove the `a` and `Shift+a` bindings and clean up the code from handling `antialiasing prev/next`.
If someone needs a switch in runtime, they can always add `antialiasing ...` to the config (but I doubt that anyone other than developers needs this).

src/config.c
```diff
@@ -37,7 +37,8 @@ static const struct config_default defaults[] = {
     { CFG_VIEWER,       CFG_VIEW_TRANSP,    "grid"                   },
     { CFG_VIEWER,       CFG_VIEW_SCALE,     "optimal"                },
     { CFG_VIEWER,       CFG_VIEW_POSITION,  "center"                 },
-    { CFG_VIEWER,       CFG_VIEW_AA,        "mks13"                  },
+    { CFG_VIEWER,       CFG_VIEW_AA_ON,     "mks13"                  },
+    { CFG_VIEWER,       CFG_VIEW_AA_START,  "yes"                    },
```
artemsen, line 41: Now we have inconsistent settings: I can set aliasing to `none` and enable it. Looks slightly confusing.

artemsen:
During the refactoring process I did it a little differently: no more `prev/next`, the `a` key toggle antialisig on/off.

voidbert:
> During the refactoring process I did it a little differently: no more `prev/next`, the `a` key toggle antialisig on/off.

Makes sense. Initially, I thought about having the option to enable/disable it during initialization, because different people may want different options (e.g.: pixel artists may want it disabled by default). I just kept prev/next for compatibility, but removing it would also be fine.

PR #311, opened 2025-06-25
Fix out-of-memory error code from shellcmd.c read_stream function

fis:
The `read_stream` function is supposed to return a negative error code on failure, and it already correctly returns `-errno` on read errors, but it can also return (the positive value) `ENOMEM` if memory allocation failed for storing the output. Returning `-ENOMEM` instead to be consistent and make those errors distinguishable.


artemsen:
Thank you!

PR #316, opened 2025-06-29
viewer: make mouse drag proportional to scale

Kuchteq:
When dragging on a hidpi scaled display, the dragged distance is not proportional to the picture's movement like it is on non scaled monitors.

https://github.com/user-attachments/assets/aabeb607-13e9-4039-9cbe-6e989d34d617


artemsen:
Come on, don't stop! =)
It's really good fix, you just need to reformat the code with clang-format and add "Signed-off-by" to commit message.
P.S. You can enable github actions in your fork to check the build in CI.

src/ui.c
```diff
@@ -269,8 +269,10 @@ static void on_pointer_motion(void* data, struct wl_pointer* wl_pointer,
                               uint32_t time, wl_fixed_t surface_x,
                               wl_fixed_t surface_y)
 {
-    const int x = wl_fixed_to_int(surface_x);
-    const int y = wl_fixed_to_int(surface_y);
+    const int x =
+        wl_fixed_to_int(surface_x * ctx.wnd.scale / FRACTION_SCALE_DEN);
```
artemsen: I believe we can calculate `ctx.wnd.scale / FRACTION_SCALE_DEN` only once to a temporary variable.

Kuchteq: good tip!

src/ui.c
```diff
@@ -269,8 +269,9 @@ static void on_pointer_motion(void* data, struct wl_pointer* wl_pointer,
                               uint32_t time, wl_fixed_t surface_x,
                               wl_fixed_t surface_y)
 {
-    const int x = wl_fixed_to_int(surface_x);
-    const int y = wl_fixed_to_int(surface_y);
+    const int scale = ctx.wnd.scale / FRACTION_SCALE_DEN;
```
artemsen: I would use double for `scale`, this will reduce the error in floating point calculations:
```
const double scale = (double)ctx.wnd.scale / FRACTION_SCALE_DEN;
```

Kuchteq: Ah, yes, of course it has to be a double, thanks for the catch! It should be good now

artemsen:
Thank you!

PR #323, opened 2025-07-07
Gallery mode by default for multiple files

sewnie:
calling swayimg for multiple files or a directory doesn't always mean that the user wants to view only the first file, instead, the user would want to view multiple files in a gallery view.

inversing the immediate mode switch to gallery for those who want to view each individual file is better because they would know they chose the correct files as well.

### DIFF
--- a/src/application.c
+++ b/src/application.c
@@ -255,6 +255,12 @@ bool app_init(const struct config* cfg, const char* const* sources, size_t num)
                                  ARRAY_SIZE(mode_names));
     ctx.mprev = mode_viewer;
 
+    // set mode to gallery if current mode is not default and
+    // the imglist is multiple images for expected usage
+    if (imglist_size() > 1 && ctx.mcurr != ctx.mprev) {
+        ctx.mcurr = mode_gallery;
+    }
+
     // create event queue notification
     ctx.event_signal = fdevent_add(handle_event_queue, NULL);
     if (ctx.event_signal < 0) {


src/application.c
```diff
@@ -255,6 +255,12 @@ bool app_init(const struct config* cfg, const char* const* sources, size_t num)
                                  ARRAY_SIZE(mode_names));
     ctx.mprev = mode_viewer;
 
+    // set mode to gallery if current mode is not default and
+    // the imglist is multiple images for expected usage
+    if (imglist_size() > 1 && ctx.mcurr != ctx.mprev) {
```
artemsen, line 260: This makes the `general.mode` option useless and adds a few questions:
1. How to start the `viewer` mode with multiple files?
2. How to start slide show?

Also, the image list at this moment contains a list of _files_, not _images_.
In fact, the list could contain hundreds of entries, but only one file with image data.

sewnie: Your concerns are correct. There is no way to check if the current mode was explicitly set by the user and not defaults. I have no workaround for this.

> Also, the image list at this moment contains a list of files, not images.
> In fact, the list could contain hundreds of entries, but only one file with image data.

Sorry, but what exactly is the reasoning for this? swayimg is unable to load the files that aren't images, so why are they remembered in memory?

artemsen:
> Sorry, but what exactly is the reasoning for this? swayimg is unable to load the files that aren't images, so why are they remembered in memory?

To understand whether a file is an image, we need to load and decode it. This is a resource-intensive and time-consuming operation. If you open a directory with thousands of photos, swayimg will take several minutes to launch.
That's why the original list contains all the files. As you go through the list, entries are removed if the loading attempt is unsuccessful.

sewnie:
> If you open a directory with thousands of photos ... entries are removed if the loading attempt is unsuccessful.

Got it. This PR still applies as the user would expect gallery mode out of the box since the given argument is a directory. Is there anyway to check if a configuration was made to override the default mode or by via argument? If not, i will simply use a shell wrapper.

artemsen:
> Is there anyway to check if a configuration was made to override the default mode or by via argument?

No. But even if there were, it would be useless, since it is not known in advance how many real images are in the list.

> simply use a shell wrapper

I'm afraid this is the only way in your case.

sewnie:
Oh well! Thanks for your time :>

PR #327, opened 2025-07-10
Add option to change info text padding

sewnie:
Adds new option: "info.padding", to change the previously hard-coded text padding within info view.

Other hard-coded values should also be taken into consideration. If those should be added as options, I'm more than willing to also contribute those. 

Hard-coded values aren't nice :<

### DIFF
--- a/src/info.c
+++ b/src/info.c
@@ -56,9 +56,6 @@ static const char* position_names[] = {
 // Max number of lines in one positioned block
 #define MAX_LINES (ARRAY_SIZE(field_names) + 10 /* EXIF and duplicates */)
 
-// Space between text layout and window edge
-#define TEXT_PADDING 10
-
 /** Key/value text surface. */
 struct keyval {
     struct text_surface key;
@@ -87,7 +84,8 @@ struct timeout {
 
 /** Info data context. */
 struct info_context {
-    bool show; ///< Show/hide flag
+    bool show;      ///< Show/hide flag
+    size_t padding; ///< Text padding
 
     struct timeout info;   ///< Text info timeout
     struct timeout status; ///< Status message timeout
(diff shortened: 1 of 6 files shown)


extra/swayimg.1
```diff
@@ -5,7 +5,7 @@
 .nh
 .ad l
 .\" Begin generated content:
-.TH "SWAYIMG" "1" "2025-06-29" "swayimg" "Swayimg manual"
+.TH "SWAYIMG" "1" "2025-07-10" "swayimg" "Swayimg manual"
```
artemsen: This changes is not necessary.

src/info.c
```diff
@@ -88,6 +86,7 @@ struct timeout {
 /** Info data context. */
 struct info_context {
     bool show; ///< Show/hide flag
+    size_t padding; ///< Text padding
```
artemsen, line 88: Please, use clang-format to [calm down CI](https://github.com/artemsen/swayimg/actions/runs/16203914844/job/45749443452#step:12:12).
You can enable actions in your fork to check CI before PR.

artemsen (changes requested):

extra/swayimg.1
sewnie: Generation of documentation. scdoc does it automatically.

artemsen: > scdoc does it automatically.

This does not make it necessary to change the timestamp without changing the content.
Please, exclude this file from PR.

sewnie: I thought this comment was referencing swayimgrc.5, sorry about that.

artemsen:
Thank you!

> Other hard-coded values should also be taken into consideration. If those should be added as options, I'm more than willing to also contribute those.

If you want to practice, there are also [hardcoded size and colors of background grid](https://github.com/artemsen/swayimg/blob/b39e37d2431566e19c3221ebe4afd11480881d48/src/viewport.c#L27).

Issue #341, opened 2025-08-19
Support zooming into mouse cursor position


JosefSchoenberger:
Thanks for developing swayimg, it is by far the best image viewer I have seen for Sway so far. There's just one minor feature that I miss.

I like zooming into my images by hovering the mouse cursor over the part of the image that I want to zoom into and turning the mouse wheel, similar to Google Maps and the like. I can already map the scrolling actions to zooming (`ScrollUp = zoom +5`/`ScrollDown = zoom -5`), but this only zooms into the center of the window, not the cursor position, so I need to drag the image again later.

Perhaps the zoom command could take another optional argument that is either `center` or `cursor` to specify the center of the zoom command? That way, key bindings could keep zooming into the center as expected (`Plus = zoom +10 center`), but the mouse wheel can zoom into the cursor position (`ScrollUp = zoom +5 cursor`).

PR #342, opened 2025-08-23
Add zoom into mouse position

JosefSchoenberger:
See #341.

This PR adds an additional optional parameter to the `zoom` action that is either `center` (with the previous behavior) or `cursor`, with which swayimg zooms into or out of the mouse cursor position, which can be better when scrolling with the mouse wheel.


src/viewport.h
```diff
@@ -125,11 +125,20 @@ bool viewport_scale_def(struct viewport* vp, const char* scale);
 const char* viewport_scale_switch(struct viewport* vp);
 
 /**
- * Set absolute scale of the image.
+ * Set absolute scale of the image, zooming into / out of the mouse position.
  * @param vp viewport context
  * @param scale scale factor (1.0 = 100%)
  */
-void viewport_scale_abs(struct viewport* vp, double scale);
```
artemsen, line 132: It would be better to have one function for scaling:
```c
void viewport_scale_abs(struct viewport* vp, double scale, size_t center_x, size_t center_y);
```
The caller should specify the position according to the action parameters.

src/viewer.c
```diff
@@ -401,27 +403,53 @@ static void zoom_image(const char* params)
     } else if (viewport_scale_def(&ctx.vp, params)) {
         name = params;
     } else {
+        const char* second_param = params;
```
artemsen: Please, replace this parsing with [str_split()](https://github.com/artemsen/swayimg/blob/6b3539b1c312726ea4317bd02ae99d21569ee2ce/src/array.h#L117), here is [example](https://github.com/artemsen/swayimg/blob/6b3539b1c312726ea4317bd02ae99d21569ee2ce/src/keybind.c#L43).

src/ui/ui.h
```diff
@@ -91,6 +91,18 @@ size_t ui_get_width(void);
  */
 size_t ui_get_height(void);
 
+/**
```
artemsen: I don't think it's a good idea to extend UI interface.
These functions are used by the viewer mode only and they are not applicable for DRM mode at all.
Viewer already has the mouse [handler](https://github.com/artemsen/swayimg/blob/6b3539b1c312726ea4317bd02ae99d21569ee2ce/src/viewer.c#L490) that has mouse coordinates, we could save them in the context. This will allow us not to change ui/wayland/drm.

src/viewer.c
JosefSchoenberger: `str_split` only takes a single delimiter, but I wanted to delimit on `isspace()` so that e.g. a tab becomes a valid parameter separator. Should I use `str_split` on spaces regardless or add a `str_split_space()` function to `src/array.c`?

src/viewport.h
JosefSchoenberger: Fixed. I had to rename the existing `center_x` and `center_y` variables in `viewport_scale_abs` to `img_center_x` and `img_center_y` for this, do these new names still make sense to you?

src/viewer.c
JosefSchoenberger: I went with the former option for new. I can trivially redo this later.

src/ui/ui.h
JosefSchoenberger: Sure, fixed. While I personally dislike keeping multiple copies of state, this has a nice benefit: If the mouse has not moved yet and so we don't know the mouse position, we now zoom into the center instead of the top left corner, which is a bit better, I guess.

JosefSchoenberger:
I integrated your feedback. While working on it, I noticed that there were rounding errors such that continuously zooming in and out without moving the cursor slightly moved the image to the top left. I changed the type of the `x` and `y` fields in the viewport context from `ssize_t` to `double`, which fixes this.
These are always rounded to the nearest integer while rendering in `viewport_draw`, so it has no influence on the image rendering, only on the internal housekeeping of the state. If one wants to be nitpicky, I guess we can change `shift` in `viewport_rotate` to a double, too. Otherwise, I think there should be no further consequences.

src/viewer.c
```diff
@@ -13,17 +14,19 @@
...
-    struct cache* preload; ///< Preloaded images
-    pthread_t preload_tid; ///< Preload thread id
-    bool preload_active;   ///< Preload in progress flag
-    bool loop_list;        ///< Loop image list
+    struct cache* history;    ///< Recently viewed images
+    struct cache* preload;    ///< Preloaded images
+    pthread_t preload_tid;    ///< Preload thread id
+    bool preload_active;      ///< Preload in progress flag
+    bool loop_list;           ///< Loop image list
+    ssize_t mouse_x, mouse_y; ///< Last known mouse position, -1 if unknown
```
artemsen: The position of the mouse pointer is always within the window size and it cannot be negative.
We can use `SIZE_MAX` as an indicator of non-valid values.
Please, replace `ssize_t` with `size_t`.

src/viewer.c
```diff
@@ -670,6 +704,8 @@ void viewer_init(const struct config* cfg, struct mode* handlers)
     // load key bindings
     ctx.kb = keybind_load(config_section(cfg, CFG_KEYS_VIEWER));
 
+    ctx.mouse_x = ctx.mouse_y = -1;
```
artemsen: Please, split this into two lines.

src/viewport.c
```diff
@@ -377,15 +378,14 @@ const char* viewport_scale_switch(struct viewport* vp)
...
+                        double center_y)
 {
     const struct pixmap* pm = viewport_pixmap(vp);
 
-    // save center
-    const double wnd_half_w = (double)vp->width / 2.0;
-    const double wnd_half_h = (double)vp->height / 2.0;
-    const double center_x = wnd_half_w / vp->scale - vp->x / vp->scale;
-    const double center_y = wnd_half_h / vp->scale - vp->y / vp->scale;
+    // save center of image
```
artemsen, line 385: Actually, this is not a "center of image".
This is the point on the window that was centered before this PR, now it can have any coordinates.

src/viewport.h
```diff
@@ -55,7 +55,7 @@ struct viewport {
     enum vp_scale def_scale;  ///< Default image scale
 
     double scale;         ///< Scale factor of the image
-    ssize_t x, y;         ///< Image position on the window surface
+    double x, y;          ///< Image position on the window surface
```
artemsen: Looks very strange now.
The image position is tied to the window, i.e. to pixels, and must be stored as an integer value.

src/viewer.c
JosefSchoenberger: Done.

src/viewer.c
JosefSchoenberger: Sure.

src/viewport.c
JosefSchoenberger: I'm confused.

> This is the point on the window that was centered before this PR

Are you talking about `wnd_half_w` and `wnd_half_h`? Because the old parameter `center_x` (which I renamed to `img_center_x`) is the center position relative to the image space, AFAICT. Perhaps I should change the comment to "calculate center position relative to the image"?

src/viewport.h
JosefSchoenberger: I don't follow. Sure, the image position is tied to the window and the unit of `x` and `y` is pixels, but that does not mean that conceptually the image must be aligned to the pixel grid.
Of course, that would significantly complicate the rendering, so we should align the image when drawing a frame. But for the position calculations, `double` makes perfect sense in my eyes. Tracking the image position to a sub-pixel precision is necessary as otherwise, repeated zooms shift the image due to rounding issues, which is not what we want.

extra/swayimgrc
```diff
@@ -212,8 +212,8 @@ ScrollLeft = step_right 5
 ScrollRight = step_left 5
 ScrollUp = step_up 5
 ScrollDown = step_down 5
-Ctrl+ScrollUp = zoom +10
-Ctrl+ScrollDown = zoom -10
+Ctrl+ScrollUp = zoom +10 cursor
```
artemsen: May be `mouse` instead of `cursor`?
Like in source code.

src/viewer.c
```diff
@@ -401,27 +404,55 @@ static void zoom_image(const char* params)
     } else if (viewport_scale_def(&ctx.vp, params)) {
         name = params;
     } else {
+        struct str_slice param_slices[2];
+        size_t param_slice_num;
+
+        param_slice_num =
```
artemsen: Please move this line down so that the declarations come first and then the code.

src/viewer.c
```diff
@@ -401,27 +404,55 @@ static void zoom_image(const char* params)
     } else if (viewport_scale_def(&ctx.vp, params)) {
         name = params;
     } else {
+        struct str_slice param_slices[2];
```
artemsen: I would use short name, for example `ps` and `ps_num`, this will make the source code more readable.

src/viewport.c
artemsen: I think the problem is in the name of the parameters =)
`center_x` is not the center. It used to be the center, but now it doesn't have to be, it can be mouse pointer coordinates.
So, `img center_x` is not the center of the image, but some point projected onto the window coordinates.
Lets name it `preserve_x` and `img_preserve_x` instead of `center_x` and `img_center_x`.

src/viewport.h
artemsen: > Tracking the image position to a sub-pixel precision

I agree that this is a good idea, but then we also need to properly handle double precision math, for example [here](https://github.com/artemsen/swayimg/blob/6b3539b1c312726ea4317bd02ae99d21569ee2ce/src/viewport.c#L76) and [here](https://github.com/artemsen/swayimg/blob/master/src/viewport.c#L350). We need to change the interface of some functions that still use `size_t`, such as [viewport_move](https://github.com/artemsen/swayimg/blob/6b3539b1c312726ea4317bd02ae99d21569ee2ce/src/viewport.c#L325C6-L325C19).

So if doubles ​​are really necessary, let's do it in another PR, because checking all the corner cases might take some time.

JosefSchoenberger: Reverted the change to `double`. I'll see if I can get to that at some point. For future reference, I also noticed that moving to the right or bottom edge does not work with `double` as well, it currently wraps around to the beginning. Just something that should be looked at once we commit to the change.

src/viewport.h
```diff
@@ -125,11 +125,14 @@ bool viewport_scale_def(struct viewport* vp, const char* scale);
...
 /**
- * Set absolute scale of the image.
+ * Set absolute scale of the image, zooming into / out of the center position.
  * @param vp viewport context
  * @param scale scale factor (1.0 = 100%)
+ * @param preserve_x the x position in viewport to zoom into / out of (in pixel)
+ * @param preserve_y the y position in viewport to zoom into / out of (in pixel)
  */
-void viewport_scale_abs(struct viewport* vp, double scale);
+void viewport_scale_abs(struct viewport* vp, double scale, double preserve_x,
```
artemsen: The coordinates here must be integers. We need to be consistent: either redesign the entire interface to support subpixels, or use integer values ​​everywhere.

src/viewer.c
```diff
@@ -401,27 +404,49 @@ static void zoom_image(const char* params)
...
+            if (str_to_num(ps[0].value, ps[0].len, &percent, 0) &&
+                percent > 0 && percent < 100000) {
                 scale = (double)percent / 100;
             }
         }
 
         if (scale != 0) {
-            viewport_scale_abs(&ctx.vp, scale);
-            info_update(info_scale, "%.0f%%", ctx.vp.scale * 100);
+            double center_x = -1, center_y = -1;
```
artemsen: 1. This is not a center, it can be a mouse pointer at any coordinates.
2. Now we have two different way to specify invalid coordinates: `-1` and `SSIZE_MAX`. It would be better to have only one.

It believe we can use `size_t` here.

src/viewport.h
JosefSchoenberger: That cast to double in `viewport.c`, line 405 is not useful either… removing it now.

src/viewer.c
JosefSchoenberger: Previously, the calculation of the center position (`wnd_half_w`) used to be as a double, too, so I kept `preserve_*` a double to retain that. Anyway, it's a `size_t` now.

artemsen:
Looks good now, thank you!

PR #344, opened 2025-09-04
Add action for setting image position

duxovni:
I find it useful to adjust the image position on the fly in the same way it's possible to adjust the scale with the `zoom` action.


src/viewport.c
```diff
@@ -365,6 +365,17 @@ bool viewport_scale_def(struct viewport* vp, const char* scale)
     return rc;
 }
 
+bool viewport_position_def(struct viewport* vp, const char* position)
+{
+    const ssize_t index = str_index(position_names, position, 0);
+    const bool rc = (index >= 0);
+    if (rc) {
```
artemsen: Maybe we should treat the empty parameter as "switch to next position"?

artemsen (commented):
Please, add description for the new action to [manual](https://github.com/artemsen/swayimg/blob/9cc13bd9c5cc0354c1f01ffa2d430d7451ff91ea/extra/swayimgrc.5.scd?plain=1#L343).
Also, it would be nice to have some default [key binding](https://github.com/artemsen/swayimg/blob/9cc13bd9c5cc0354c1f01ffa2d430d7451ff91ea/src/config.c#L115).

duxovni:
Good points, thank you!

src/viewport.c
```diff
@@ -407,6 +407,28 @@ void viewport_scale_abs(struct viewport* vp, double scale, size_t preserve_x,
     fixup_position(vp, false);
 }
 
+bool viewport_position_def(struct viewport* vp, const char* position)
```
artemsen: Lets merge these two functions. If `position` is NULL, then switch to the next position.

src/config.c
```diff
@@ -144,6 +144,16 @@ static const struct configdef_kv def_keysv[] = {
     { "BackSpace",        "zoom optimal"                             },
     { "k",                "zoom keep"                                },
     { "Alt+s",            "zoom"                                     },
+    { "Alt+period",       "position center"                          },
```
artemsen: Sorry, I didn't express myself quite correctly.
By default key binding I meant one method used as an example. No one will remember this whole set of keys anyway and will make their own.
Let's leave only `Alt+period=position` to switch the position to next value.

src/viewport.c
duxovni: Good idea; thanks!

src/config.c
duxovni: Fair enough haha

PR #361, opened 2025-12-30
Improve gallery options

proxict:
First of all, thank you for this very nifty image viewer!

This pull request has relatively many changes, but it's split into hopefully atomic enough commits for easier review.

Here is a list of the functional changes in this PR:
 - Border width made configurable with `border_width` config option
 - Selected thumbnail scale made configurable with `selected_scale` config option
 - Thumbnail padding made configurable with `padding` config option
 - Fixed thumbnail centering
   - thumbnails are now rendered in the center of the window even after changing the thumbnail size
 - Added option to keep the aspect ratio of thumbnails instead of forcing them to squares (`thumb_scale` config option)
   - border and shadow are adjusted to fit the real thumbnail's aspect ratio

Regarding the `thumb_scale` config option, it replaces the `fill` option. I wasn't sure about the best way of going about it and decided to deprecate the `fill` option in favor of the `thumb_scale` one. It's of course possible to keep both options with one taking precedence over the other or any other solution that you deem most suitable.
Now, a deprecation warning is printed when the `fill` option is used, and a message suggesting the new `thumb_scale` option is printed.

Preview:
image
image

### DIFF
--- a/src/application.c
+++ b/src/application.c
@@ -11,7 +11,6 @@
 #include "imglist.h"
 #include "info.h"
 #include "ipc.h"
-#include "shellcmd.h"
 #include "slideshow.h"
 #include "tpool.h"
 #include "ui/ui.h"
--- a/src/application.h
+++ b/src/application.h
@@ -4,9 +4,12 @@
 
 #pragma once
 
+#include "action.h"
+#include "config.h"
 #include "fs.h"
 #include "image.h"
-#include "keybind.h"
+
+#include <xkbcommon/xkbcommon.h>
 
 /**
  * Handler of the fd poll events.
--- a/src/info.h
+++ b/src/info.h
@@ -5,7 +5,7 @@
 #pragma once
 
 #include "image.h"
-#include "mode.h"
+#include "keybind.h"
 
 /** Available info fields. */
 enum info_field {
--- a/src/keybind.c
+++ b/src/keybind.c
@@ -4,7 +4,7 @@
 
 #include "keybind.h"
 
-#include "application.h"
+#include "array.h"
 
 #include <stdlib.h>
 #include <string.h>
--- a/src/slideshow.c
+++ b/src/slideshow.c
@@ -8,7 +8,6 @@
 #include "fdpoll.h"
 #include "imglist.h"
 #include "info.h"
-#include "render.h"
 #include "tpool.h"
 #include "ui/ui.h"
 #include "viewport.h"
--- a/src/ui/drm.c
+++ b/src/ui/drm.c
@@ -2,7 +2,7 @@
 // DRM based user interface.
 // Copyright (C) 2025 Artem Senichev <artemsen@gmail.com>
 
-#include "../application.h"
+#include "../config.h"
 #include "buildcfg.h"
 #include "uiface.h"
 
--- a/src/ui/wayland.c
+++ b/src/ui/wayland.c
@@ -5,6 +5,7 @@
 #include "../application.h"
 #include "../array.h"
 #include "../fdpoll.h"
+#include "../keybind.h"
 #include "uiface.h"
 #include "wndbuf.h"
 


extra/swayimgrc
```diff
@@ -96,8 +96,8 @@ cache = 100
 preload = no
 # Enable/disable storing thumbnails in persistent storage (yes/no)
 pstore = no
-# Fill the entire tile with thumbnail (yes/no)
-fill = yes
+# Thumbnail scale mode (fit/fill/real)
```
artemsen: The "real scale" variant is confusing. I would have thought it meant the thumbnail would be the same size as the image.
Perhaps it's not the scale, but the aspect ratio. Let's rename `thumb_scale` to `aspect`?

src/image.h
```diff
@@ -98,6 +98,15 @@ enum image_status {
     imgload_unknown      ///< Unknown errors
 };
 
+/** Thumbnail scale mode. */
+enum thumb_scale_mode {
```
artemsen: `thumb_aspect` would be more clear.

src/config.c
```diff
@@ -227,6 +227,19 @@ static const struct configdef_kv def_keysg[] = {
     { "Ctrl+ScrollDown",  "thumb -20"                                },
     { "MouseLeft",        "mode viewer"                              },
 };
+
+// clang-format on
+/** Deprecated configuration options. */
```
artemsen: I appreciate your concern for users, but I think this is unnecessary.
It would be nice to provide a warning and convert the settings to the new format, but I don't like this legacy support situation. Besides, it's unclear when the warning can be removed (next release? in 1 moth or year?).

src/pixmap.c
```diff
@@ -456,38 +456,45 @@ void pixmap_blend(struct pixmap* pm, ssize_t x, ssize_t y, size_t width,
...
+                  size_t thickness, argb_t color)
 {
-    if (y >= 0 && y < (ssize_t)pm->height) {
-        const ssize_t begin = max(0, x);
-        const ssize_t end = min((ssize_t)pm->width, x + (ssize_t)width);
-        const size_t offset = y * pm->width;
-        for (ssize_t i = begin; i < end; ++i) {
-            pixmap_alpha_blend(color, &pm->data[offset + i]);
+    for (size_t j = 0; j < thickness; ++j) {
+        ssize_t yy = y + j;
```
artemsen: Can be `const`

src/pixmap.c
```diff
@@ -456,38 +456,45 @@ void pixmap_blend(struct pixmap* pm, ssize_t x, ssize_t y, size_t width,
...
-                  argb_t color)
+                  size_t thickness, argb_t color)
 {
-    if (x >= 0 && x < (ssize_t)pm->width) {
-        const ssize_t begin = max(0, y);
-        const ssize_t end = min((ssize_t)pm->height, y + (ssize_t)height);
-        for (ssize_t i = begin; i < end; ++i) {
-            pixmap_alpha_blend(color, &pm->data[i * pm->width + x]);
+    for (size_t j = 0; j < thickness; ++j) {
+        ssize_t xx = x + j;
```
artemsen: `const`

extra/swayimgrc
proxict: Good call, `aspect` is definitely more fitting. Renamed `thumb_scale` to `aspect`.

src/config.c
proxict: I wasn't sure how important backwards compatibility with the config file is so I took what I thought was a safer route, but I agree that maintaining deprecations is painful.  
I've removed the warning and the `fill` option is as if it had never existed.

proxict:
Thank you for the prompt review!

I have, hopefully, incorporated all the review notes.

artemsen:
Thank you!

proxict:
My pleasure!

artemsen:
oops.. Looks like we lost background color for transparent images.
Old version:
image
New version:
image

proxict:
Oh, silly me, I forgot about transparency!
The problem is [here](https://github.com/artemsen/swayimg/pull/361/commits/370c72a8f65596de3e6c41c6e0245b547db02cac#diff-e0027f1bef361126c321890557f061473c9c651fff317ed72d17b8a6f4de6443R437-R440). I'll fix it and create a new PR.

Issue #364, opened 2026-01-03
hyprland: overlay config option has no effect


kikeijuu:
**Describe the bug**
Setting the `overlay` config option to `yes` in swayimg's config file or as a command flag does not result in expected behavior, on endeavorOS with hyprland.

**To Reproduce**
1. Install `swayimg` or `swayimg-git`
2. Set `overlay = yes` in general settings
3. Open any image, eg. with `swayimg Pictures/flandre.png`
4. Cry, because the overlay feature is why you adopted swayimg ;-;

**Expected behavior**
The window to spawn on top of the terminal and conform to it's boundaries, acting as a visual overlay.

**Screenshots and examples**

Here, swayimg tiles instead of becoming an overlay of the bottom left terminal. Swayimg config and hyprland windowrules also pictured.

<img width="1918" height="1198" alt="Image" src="https://github.com/user-attachments/assets/162c0a48-cff7-415a-b57b-ae76e0212321" />

**Version:**
4.6

**Additional context**
The overlay definitely worked before the change was made that stopped it from being default. As for after that, I am unsure. It may be that specific windowrules should be set in hyprland's config to ensure that swayimg can do it's overlay thing. If this is the case, pls add them to the readme qwq

There is another discussion thread that has windowrule 'tips' for swayimg, but they are to use it in a very different way, explicitly not as an overlay. For example, making swayimg float just deposits it awkwardly in the center of the screen, and it does not act as an overlay.

I really enjoy using swayimg as my img viewer in Yazi, as it allows me to view images without disrupting my tiling or interrupting my tui file manager.

I can provide more information if that would help, and I have tested this with clean reinstalls of swayimg, --chroot builds, and new configs. The issue persists across these, unfortunately.

Thank you so much for swayimg, it is a beautiful tool ^w^

artemsen:
I'm pretty sure it's due to breaking changes in the window rules format in the latest version of Hyprland.
I don't use this WM and unfortunately I lost my virtual machine with Hyprland, so fixing these rules may take a long time.
If you need overlay as soon as possible, you can fix rules [here](https://github.com/artemsen/swayimg/blob/ab86ee4d4ce2a51e7f3487e2a4bbc6076d9cb21b/src/ui/compositor.c#L271). Pull requests are welcome =)

kikeijuu:
Ah right, those gave my own config quite a lot of trouble too the other day >.>

I'll see what I can do!

kikeijuu:
Making progress ^^

[...]

PR #366, opened 2026-01-03
attempted fix of windowrule syntax for hyprland

kikeijuu:
image

So far, this PR changes the syntax of windowrules in the compositor.c file (just two lines!) to match the latest hyprland release. This allows the automatic placement and floating overlay behavior of swayimg to be restored, for users who enable it.

However, something is wrong with the sizing. As you can see in the image, swayimg is consistently half the size it should be, with slight overhangs. My theory is that

> It may be that the dimensions of swayimg itself are being passed back into swayimg when it is deciding its sizing, if somehow the window is being spawned before dimensions are calculated, but I understand far too little of the two compositor files to say so definitely <.<

But as I say here I am not certain >.<

At any rate, this is less of a regression than the current state of swayimg, where overlays do not occur at all due to the syntax changes in hyprland: see https://github.com/artemsen/swayimg/issues/364 for more details.

Due to the above, while this PR should be fine on its own, help figuring out what is causing either

a) the `compositor.c` function to grab hyprland client dimensions
b) the `compositor.h` function to resize the swayimg window

to function improperly (at least I think that is the case) on hyprland, or test the PR on their own machine, that would be awesome ^^

(I am still very new to github, sorry if any of this is improper)

### DIFF
--- a/src/ui/compositor.c
+++ b/src/ui/compositor.c
@@ -298,6 +298,15 @@ static bool hyprland_overlay(const struct wndrect* wnd, char** app_id)
     }
     json_object_put(response);
 
+    // set size
+    snprintf(buf, sizeof(buf), "keyword windowrule size %zd %zd, match:class %s",
+            wnd->width, wnd->height, *app_id);
+    response = hyprland_request(buf);
+    if (!response) {
+        return false;
+    }
+    json_object_put(response);
+
     return true;
 }
 


artemsen:
Sorry, I can't accept this PR.
It partially solves the problem, but the overlay functionality still doesn't work as it should.

kikeijuu:
Odd, I definitely added a sign off using the formatting, in the second commit. Sorry, I'm still very new to this >.>

kikeijuu:
> Sorry, I can't accept this PR. It partially solves the problem, but the overlay functionality still doesn't work as it should.

Overlay functionality now works as it should! works with fractional scaling on or off, did a few tests. It would be good to get another hyprland user to test but I think this is now good to go, commit sign-offs aside (sorry <.<)

image

artemsen:
> Odd, I definitely added a sign off using the formatting, in the second commit.

It should be in all commits.
It doesn't matter, I squashed them anyway.
Thank you!

PR #368, opened 2026-01-05
Handle CMYK JPEGs with quick-and-dirty colorspace conversion

### LINKED ISSUE #367: swayimg fails to open CMYK JPEG images
**Describe the bug**
When I try to open a JPEG that uses the CMYK colorspace, swayimg exits with "Invalid format".

**To Reproduce**
Attempt to open a CMYK JPEG in swayimg.

**Expected behavior**
swayimg should load and display the image with the correct colors.

**Screenshots and examples**
Here's a sample JPEG using the CMYK colorspace:
Peppers in CMYK

**Version:**
4.6

**Additional context:**
Prior to commit 08f28204, swayimg could open CMYK JPEGs but displayed them with the colors messed up (you can see some of my desktop background showing through the translucent areas):
Peppers with wonky colors

duxovni:
Resolves #367.

Screenshot of swayimg with this patch displaying the CMYK peppers image from #367:
image

### DIFF
--- a/src/formats/jpeg.c
+++ b/src/formats/jpeg.c
@@ -55,7 +55,17 @@ enum image_status decode_jpeg(struct imgdata* img, const uint8_t* data,
     jpeg_read_header(&jpg, TRUE);
 
 #ifdef LIBJPEG_TURBO_VERSION
-    jpg.out_color_space = JCS_EXT_BGRA;
+    switch (jpg.jpeg_color_space) {
+        case JCS_CMYK:
+        case JCS_YCCK:
+            jpg.out_color_space = JCS_CMYK;
+            break;
+        case JCS_UNKNOWN:
+            break;
+        default:
+            jpg.out_color_space = JCS_EXT_BGRA;
+            break;
+    }
 #endif // LIBJPEG_TURBO_VERSION
     jpeg_start_decompress(&jpg);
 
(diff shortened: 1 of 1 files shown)


src/formats/jpeg.c
```diff
@@ -55,7 +55,17 @@ enum image_status decode_jpeg(struct imgdata* img, const uint8_t* data,
     jpeg_read_header(&jpg, TRUE);
 
 #ifdef LIBJPEG_TURBO_VERSION
-    jpg.out_color_space = JCS_EXT_BGRA;
+    switch (jpg.jpeg_color_space) {
+        case JCS_CMYK:
+        case JCS_YCCK:
```
artemsen, line 60: What about `JCS_YCbCr`?

src/formats/jpeg.c
```diff
@@ -55,7 +55,17 @@ enum image_status decode_jpeg(struct imgdata* img, const uint8_t* data,
...
+            jpg.out_color_space = JCS_CMYK;
+            break;
+        case JCS_UNKNOWN:
```
artemsen, line 63: Looks suspicious: now we have undefined `jpg.out_color_space`.
Lets remove handling `JCS_UNKNOWN`.

src/formats/jpeg.c
```diff
@@ -70,18 +80,33 @@ enum image_status decode_jpeg(struct imgdata* img, const uint8_t* data,
...
                 const uint8_t* src = line + x * jpg.out_color_components;
-                switch (jpg.out_color_components) {
-                    case 1:
-                        pixel[x] = ARGB(ARGB_MAX_COLOR, src[0], src[0], src[0]);
-                        break;
-                    case 3:
-                        pixel[x] = ARGB(ARGB_MAX_COLOR, src[0], src[1], src[2]);
-                        break;
+                if (jpg.out_color_space == JCS_CMYK) {
+                    double c = src[0];
```
artemsen: Can be `const`.

src/formats/jpeg.c
```diff
@@ -70,18 +80,33 @@ enum image_status decode_jpeg(struct imgdata* img, const uint8_t* data,
...
+                    double m = src[1];
+                    double y = src[2];
+                    double k = src[3];
+
+                    uint8_t r = (c * k / ARGB_MAX_COLOR);
```
artemsen: 1. Can be `const`;
2. Unnecessary brackets (do you mean `(m * k) / ARGB_MAX_COLOR` ?).

artemsen (changes requested):

src/formats/jpeg.c
duxovni: libjpeg-turbo supports converting `JCS_YCbCr` to `JCS_EXT_BGRA` automatically:
https://raw.githubusercontent.com/libjpeg-turbo/libjpeg-turbo/main/doc/libjpeg.txt
```
jdcolor.c currently supports
        YCbCr => RGB
        YCbCr => GRAYSCALE
        RGB => GRAYSCALE
        GRAYSCALE => RGB
        YCCK => CMYK
```

src/formats/jpeg.c
duxovni: It's not undefined:
https://raw.githubusercontent.com/libjpeg-turbo/libjpeg-turbo/main/doc/libjpeg.txt
```
J_COLOR_SPACE out_color_space
        Output color space.  jpeg_read_header() sets an appropriate default
        based on jpeg_color_space
```
If we set it to `JCS_EXT_BGRA`, then swayimg will fail to display the image at all. If we leave it as its default value, then libjpeg won't attempt any colorspace conversion, and swayimg will assume it's grayscale, RGB, or RGBA based on the number of components, so it will still display the image and worst-case the colors will look weird, which is the best we can do since we don't know the colorspace.

src/formats/jpeg.c
duxovni: Ah, yup, good catch; the brackets are left over from when I was doing explicit casting rather than just letting the type declarations handle it.

PR #370, opened 2026-01-06
info: add toggle mode cycling with customizable mode lists

mbroemme:
## Description
Extend info to support support comma-separated mode cycling.

* `info` (no expression) cycles through the default list: `off/viewer/slideshow/gallery`
* `info MODE` switches directly to the specified mode
* `info MODE,MODE,...` cycles through the provided comma-separated mode list

Documentation has been updated to reflect the new syntax and examples.

## Use Case
Press `i` to show minimal info in viewer, and hide it everywhere else, you want:

* In viewer mode: `i` toggles info overlay between `viewer` and `off` (so you can quickly peek at filename/size/etc.).
* In gallery mode: `i` toggles between `gallery` and `off` (so it doesn't switch you into the viewer overlay style).

```
[keys.viewer]
i = info viewer,off

[keys.gallery]
i = info gallery,off
```

Result: The same key (`i`) acts as a context-aware "info overlay toggle" depending on where you are, without cycling through modes you don’t want.

### DIFF
--- a/src/info.c
+++ b/src/info.c
@@ -576,117 +576,111 @@ void info_destroy(void)
 
 void info_switch(const char* expression)
 {
-    const char* toggle_kw = "toggle";
-    const size_t toggle_kw_len = strlen(toggle_kw);
-    struct str_slice args[2];
-    size_t argc = 0;
-    bool handled = false;
+    struct str_slice slices[6];
+    size_t num = 0;
+    bool cycle = false;
 
     timeout_reset(&ctx.info);
 
-    if (expression) {
-        argc = str_split(expression, ' ', args, ARRAY_SIZE(args));
+    if (!expression || !*expression) {
+
+        // default cycle order
+        expression = "off,viewer,slideshow,gallery";
+        cycle = true;
     }
 
-    if (argc > 0) {
+    // comma-separated list: MODE,MODE,...
+    num = str_split(expression, ',', slices, ARRAY_SIZE(slices));
+    if (num == 0) {
 
-        // toggle in a custom list: "toggle [mode1/mode2/...]"
-        if (args[0].len == toggle_kw_len &&
-            strncmp(args[0].value, toggle_kw, toggle_kw_len) == 0) {
-            const char* spec =
-                argc > 1 && args[1].len ? args[1].value : "off/viewer/gallery";
-            struct str_slice slices[16];
-            size_t num, start;
-            const char* current_name;
-            size_t current_len;
-            size_t current_idx = SIZE_MAX;
-            bool switched = false;
+        // expression contains only spaces
+        expression = "off,viewer,slideshow,gallery";
+        cycle = true;
+        num = str_split(expression, ',', slices, ARRAY_SIZE(slices));
+    }
+    if (num > ARRAY_SIZE(slices)) {
+        num = ARRAY_SIZE(slices);
+    }
+    if (num > 1) {
+        cycle = true;
+    }
 
-            handled = true;
+    if (!cycle) {
 
-            // split mode list (spaces are stripped by str_split)
-            num = str_split(spec, '/', slices, ARRAY_SIZE(slices));
-            if (num > ARRAY_SIZE(slices)) {
-                num = ARRAY_SIZE(slices);
-            }
+        // explicit mode
+        const char* mode = slices[0].value;
+        const size_t len = slices[0].len;
 
-            // detect current mode ("off" if hidden)
-            if (!ctx.show) {
-                current_name = "off";
-            } else if (ctx.current) {
-                current_name = ctx.current->name;
+        if (len != 0) {
+            if (len == 3 && strncmp(mode, "off", 3) == 0) {
+                ctx.show = false;
             } else {
-                current_name = "";
-            }
-            current_len = strlen(current_name);
-
-            // find current index
-            for (size_t index = 0; index < num; ++index) {
-                const char* mode = slices[index].value;
-                const size_t len = slices[index].len;
-
-                if (len == current_len &&
-                    strncmp(mode, current_name, current_len) == 0) {
-                    current_idx = index;
-                    break;
-                }
-            }
-
-            // select next non-empty entry
-            start = current_idx != SIZE_MAX ? current_idx + 1 : 0;
-            for (size_t step = 0; step < num && !switched; ++step) {
-                const size_t idx = (start + step) % num;
-                const char* s = slices[idx].value;
-                const size_t len = slices[idx].len;
-
-                if (len == 0) {
-                    continue;
-                }
-
-                if (len == 3 && strncmp(s, "off", 3) == 0) {
-                    ctx.show = false;
-                    switched = true;
-                    continue;
-                }
-
-                struct scheme* scheme = find_scheme(s, len);
+                struct scheme* scheme = find_scheme(mode, len);
                 if (scheme) {
                     ctx.current = scheme;
                     ctx.show = true;
-                    switched = true;
                 } else {
-                    fprintf(stderr, "Invalid info scheme: %.*s\n", (int)len, s);
+                    fprintf(stderr, "Invalid info scheme: %.*s\n", (int)len,
+                            mode);
                 }
             }
-        } else if (args[0].len == 3 && strncmp(args[0].value, "off", 3) == 0 &&
-                   argc == 1) {
-            ctx.show = false;
-            handled = true;
-        } else {
-
-            // Set explicit scheme (legacy behaviour).
-            char* name = malloc(args[0].len + 1);
-            if (name) {
-                memcpy(name, args[0].value, args[0].len);
-                name[args[0].len] = 0;
-                info_set_default(name);
-                free(name);
-            } else {
-                fprintf(stderr, "Out of memory\n");
-            }
-            ctx.show = true;
-            handled = true;
         }
-    }
+    } else {
+        size_t start;
+        const char* current_name;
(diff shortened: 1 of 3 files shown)


src/info.c
```diff
@@ -557,9 +592,89 @@ void info_destroy(void)
...
 void info_switch(const char* name)
 {
+    const char toggle_kw[] = "toggle";
+
     timeout_reset(&ctx.info);
 
     if (name && *name) {
+
+        // toggle in a custom list: "toggle [mode1/mode2/...]" (spaces ignored)
+        if (strncmp(name, toggle_kw, sizeof(toggle_kw) - 1) == 0 &&
```
artemsen: I believe it would be better to split `name` with `split_string` instead of writing a new parser, it strips spaces.
Also, I would rename `name` to `expression`, as it is not a name anymore.

src/info.c
```diff
@@ -557,9 +592,89 @@ void info_destroy(void)
...
+            (name[sizeof(toggle_kw) - 1] == 0 ||
+             isspace((unsigned char)name[sizeof(toggle_kw) - 1]))) {
+            const char* spec = name + sizeof(toggle_kw) - 1;
+            struct str_slice slices[16];
+            const char* current_name;
+            size_t current_len;
+            ssize_t current_idx = -1;
+            size_t num, start;
+
+            while (*spec && isspace((unsigned char)*spec)) {
```
artemsen: Just curious, why are you cast `char` to `unsigned char` in these calls?

src/info.c
```diff
@@ -557,9 +592,89 @@ void info_destroy(void)
...
+                ++spec;
+            }
+            if (!*spec) {
+                spec = "off/viewer/gallery";
```
artemsen: Strange choice of default. Why not `viewer/gallery` or `viewer/slideshow/gallery`?

src/info.c
```diff
@@ -557,9 +592,89 @@ void info_destroy(void)
...
+                current_name = ctx.current->name;
+            } else {
+                current_name = "";
+            }
+            current_len = strlen(current_name);
+
+            // find current index
+            for (size_t i = 0; i < num; ++i) {
+                const char* s = slices[i].value;
+                size_t l = slices[i].len;
```
artemsen: 1. Please, avoid such a short name for variables. Name it `len` at least.
2. Can be `const` (and similar cases in the other parts of code).

src/info.c
```diff
@@ -557,9 +592,89 @@ void info_destroy(void)
...
+                    struct scheme* scheme = find_scheme(s, l);
+                    if (scheme) {
+                        ctx.current = scheme;
+                        ctx.show = true;
+                    } else {
+                        fprintf(stderr, "Invalid info scheme: %.*s\n", (int)l,
+                                s);
+                    }
+                }
+                return;
```
artemsen: It is a bad practice to return from a middle of the function. This is acceptable in some cases, but not in this one.
You've written a long and complex parser, and it's too difficult to analyze, especially when we have two returns ​​at this point.

src/info.c
```diff
@@ -557,9 +592,89 @@ void info_destroy(void)
 
 void info_switch(const char* name)
 {
+    const char toggle_kw[] = "toggle";
```
artemsen: It's not a crime if you make a pointer to a string and a variable with its length.
It would be more readable than multiple `sizeof(toggle_kw) - 1`.

src/info.c
```diff
@@ -557,9 +592,89 @@ void info_destroy(void)
...
     timeout_reset(&ctx.info);
 
     if (name && *name) {
+
+        // toggle in a custom list: "toggle [mode1/mode2/...]" (spaces ignored)
+        if (strncmp(name, toggle_kw, sizeof(toggle_kw) - 1) == 0 &&
+            (name[sizeof(toggle_kw) - 1] == 0 ||
+             isspace((unsigned char)name[sizeof(toggle_kw) - 1]))) {
+            const char* spec = name + sizeof(toggle_kw) - 1;
+            struct str_slice slices[16];
```
artemsen: Why `16`?

src/info.c
```diff
@@ -557,9 +592,89 @@ void info_destroy(void)
...
+            const char* current_name;
+            size_t current_len;
+            ssize_t current_idx = -1;
```
artemsen: It would be better to use `ssize_t` and `SSIZE_MAX` instead of negative index.

mbroemme:
@artemsen Thanks for the feedback, will review and update the PR in the next days!

src/info.c
mbroemme: Because the `<ctype.h>` functions/macros (like `isspace()`) are only defined for values that are either `EOF` or representable as an `unsigned char`. In C, plain `char` may be signed. If it is, any byte with the high bit set (e.g. `0x80..0xFF`, which is common in UTF-8 or non-ASCII input) becomes a negative `char`. Passing a negative value (other than `EOF`) into `isspace()` is undefined behavior - many implementations index a lookup table with that value, so you can end up indexing out of bounds. In this case, it is basically extra safety.

src/info.c
mbroemme: Because "toggle" is typically bound to a single key (like `i`) and people expect that key to also hide the overlay. If the default were `viewer/gallery`, you'd need a separate bind (`info off`) to actually turn it off. I'm using only the viewer with `swayimg` and bind `i = info toggle viewer/off` to turn on/off the info overlay.

In the implementation, if the current mode isn't found in the provided list (or the list is weird/empty), it starts from the beginning. Having `off` first makes the fallback behavior "hide overlay" instead of "force an overlay on," which is the least-surprising/least-intrusive outcome.

src/info.c
mbroemme: Fixed in d1a6369b0174863d13b6caf595c6a2fd059d74c4 I haven't checked before that `str_split` exist in `src/array.c`. Using it now.

src/info.c
mbroemme: Fixed in d1a6369b0174863d13b6caf595c6a2fd059d74c4 as well. With switch to `str_split`, I'm no longer using `isspace` and don't cast explicitly.

src/info.c
mbroemme: Fixed in b9f836765098c147c57cea5ccf61de12eaa5d113

src/info.c
mbroemme: Fixed in c5c748c01b318423f9991d8f6e51a394ef8ba6b6, using pointer and removed explicit cast.

src/info.c
mbroemme: Fixed in 537d1f8cb77e9717f2a76075b7df85edf02298d8, no more mid-function returns, using `handled` and `switched` helper variables.

src/info.c
mbroemme: Fixed in c58910944610babdb176dba489f0cfc305ab9eac, using `size_t` and `SIZE_MAX` which are fine for this case and makes code more readable and avoid additional explicit cast.

src/info.c
mbroemme: It's just a hard cap to avoid dynamic allocation while parsing a user-provided toggle list. `str_split()` needs you to provide an output array; using a fixed-size stack array is cheap and predictable. The `16` means:
* up to 16 entries in `toggle`'s slash-list will be considered
* anything beyond that is ignored (because the array is full)
There's nothing magic about `16`- it's a "reasonable enough" limit for a keybind expression (people typically won't cycle through dozens of modes). If you prefer, we can make it clearer by turning it into a named constant, e.g.:
```c
#define TOGGLE_MAX_ENTRIES 16
struct str_slice slices[TOGGLE_MAX_ENTRIES];
```
or even reuse an existing project-wide constant if there is one.

mbroemme:
@artemsen Addressed your questions and incorporated your feedback. I can squash the commits if you'd like.

src/info.c
artemsen: What I wanted to say is that 16 options for 4 available modes is too much. 6 would be enough considering the understudies.

src/info.c
artemsen: I think that if a user wants a custom switching scheme, he should specify it without fail.
Moreover, I think that the key word "toggle" is redundant.
Therefore, I propose the following scheme:
- `info` (without expression): toggle off/viewer/slideshow/gallery;
- `info MODE`: switch to specified MODE;
- `info MODE,MODE,MODE,...`: switch between specified modes.

src/info.c
artemsen: This doesn't make the code more readable, but let's see where we end up =)

src/info.c
```diff
@@ -576,117 +576,111 @@ void info_destroy(void)
 
 void info_switch(const char* expression)
 {
-    const char* toggle_kw = "toggle";
-    const size_t toggle_kw_len = strlen(toggle_kw);
-    struct str_slice args[2];
-    size_t argc = 0;
-    bool handled = false;
+    struct str_slice slices[6];
```
mbroemme, line 579: Changed to 6 as discussed in https://github.com/artemsen/swayimg/pull/370#discussion_r2674914068

src/info.c
mbroemme: Changed with commit 52fa0e0296a90a41729d87ca251daacd253bc48b

src/info.c
mbroemme: @artemsen Review it again please.

artemsen:
Thank you!

To be honest, I have a few complaints, but I don't want to bother you any further.
If you don't mind, I'll optimize the code a bit myself.

Issue #221, opened 2024-11-20
config for shown exif attributes


litoj:
Sometimes I want to know the settings I used for a given photo. Currently, the exif attributes shown are hardcoded. This means for each image I want to compare I must go to the cli and find it with `exiftool`. It would be really useful if we could set the wanted attributes in the config.

artemsen:
Implemented in latest version. See [swayimg.viewer.set_text_tl](https://github.com/artemsen/swayimg/blob/master/CONFIG.md#swayimgviewerset_text_tl) for details.
For example:
```lua
swayimg.viewer.set_text_tl({
  "File: {name}",
  "EXIF date: {meta.Exif.Photo.DateTimeOriginal}",
  "EXIF camera: {meta.Exif.Image.Model}"
})
```

PR #378, opened 2026-02-05
feat(exif): allow exiftool use for configurable metadata display

### LINKED ISSUE #369: Support exif in all AVIF and JXL formats
**Is your feature request related to a problem? Please describe.**
I export all of my images into .jxl or .avif. However, JXL metadata doesn't show up no matter which software I use to convert the image to jxl. In case of AVIF, I can see the metadata if compressed using `imagemagick` but not if I used `darktable` or `gimp`, which I usually do because of photo editing.

**Describe the solution you'd like**
Support for the AVIF exif format used by GIMP and darktable. Support for JXL exif format.

**Describe alternatives you've considered**
There is no possible alternative for JXL - neither the container nor image formats show the exif data in swayimg. For AVIF I could convert all of my images using imagemagick, but that would decrease their quality.

**Additional context**
all the files produced by the individual softwares - example is a photo edited to be pure white 64x64. PNG uses xmp, so the exif shown looks just like some XML, but I personaly don't care for exif on PNG, only the photo formats.

[avif_jxl_exif.zip](https://github.com/user-attachments/files/24432003/avif_jxl_exif.zip)

litoj:
This PR resolves #369 and #221 by allowing the user to optionally use exiftool to query a list of exif parameters of his choosing replacing what was read by libexif.

### DIFF
--- a/src/exiftool.c
+++ b/src/exiftool.c
@@ -9,21 +9,21 @@
 #include <string.h>
 #include <unistd.h>
 
-static void add_if_empty(struct imgdata *data, const char* text)
+static void add_if_empty(struct imgdata* data, const char* text)
 {
     if (!data->info) {
         image_add_info(data, "Exif", "%s", text);
     }
 }
 
-void query_exiftool(const struct image *img, const char* arg_query)
+void query_exiftool(const struct image* img, const char* arg_query)
 {
     if (!img->data || (img->data->info && img->data->used_exiftool)) {
         return;
     }
     img->data->used_exiftool = true;
 
-    FILE *fp;
+    FILE* fp;
     char command[512];
     const int buf_size = 512;
     char buffer[buf_size];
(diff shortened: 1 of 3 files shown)


artemsen:
I really appreciate your efforts, but I've taken on a complete rewrite of the project.
The new version uses exiv2 and allows to [customize](https://github.com/artemsen/swayimg/blob/4d35442a8918bb9e8e825e607da1910d7931a4ad/extra/example.lua#L16) the displayed fields via lua.

litoj:
That does sound amazing and as a neovim (lua) user I can't wait to test it out.

However, apparently even exiv2 still doesn't support jxl - see https://github.com/Exiv2/exiv2/issues/1506.

Someone seems to have implemented the support it in their fork, but that was now almost 3 years ago and the issue still remains open, so I assume it is still not possible with the main version of exiv2.

This could then stand as a temporary fix until a proper C/C++ implementation is widely available.

artemsen:
> However, apparently even exiv2 still doesn't support jxl - see https://github.com/Exiv2/exiv2/issues/1506.

At least exiv2 0.28.7 from the Arch package supports JXL tags (as well as other formats from #369).

Sorry, but I don't see any practical point in merging this PR given that the current implementation will be replaced.

litoj:
> At least exiv2 0.28.7 from the Arch package supports JXL tags (as well as other formats from #369).

Really? That's great then. Though I wonder why they have that issue still open.

Anyway, in that case there really is no point to this PR.

PR #383, opened 2026-02-28
Gallery: add image marking

very-amused:
This PR adds a `mark` action in the gallery viewer (bound to `m` by default). Marked images will have their paths written to stdout, one image path per line, before swayimg exits. This feature is directly inspired by [sxiv](https://github.com/xyb3rt/sxiv)'s `-o` option which allows it to be used as a image picker in a pipe-based workflow. I've personally used this feature of sxiv for years in my X11 wallpaper setting script, so it's very helpful to have this in swayimg as I'm moving to Wayland.


artemsen:
Unfortunately, you've come at a bad time - I'm in the middle of a [major rework](https://github.com/artemsen/swayimg/tree/v5) of the project.
Changes merged into the master branch will be lost, so let's pause this PR and come back to it a little later after moving to the new version of the application.

very-amused:
Sounds good. If I may ask: what's motivating the switch from C to C++ in v5?

artemsen:
> Sounds good. If I may ask: what's motivating the switch from C to C++ in v5?

The project grew significantly, and it became difficult to reinvent arrays, strings, algorithms, etc. Furthermore, manual memory management became quite tedious.

artemsen:
The latest version of Swayimg already [has this feature](https://github.com/artemsen/swayimg/blob/master/CONFIG.md#swayimgimagelistmark), moreover it works in viewer/slideshow mode too.

PR #384, opened 2026-02-28
Add explicit mode prev action

very-amused:
This PR adds an explicit `mode prev` action that achieves the same as `mode` with no param does currently. This form is less vague/unclear than just using `mode` to switch to the previous viewer mode, but the existing syntax will still continue to work.

### DIFF
--- a/src/application.c
+++ b/src/application.c
@@ -66,7 +66,8 @@ void app_switch_mode(const char* name)
         const ssize_t index = str_index(mode_names, name, 0);
         if (index >= 0) {
             next = index;
-					// keep compatibility w/ existing behavior of `mode` (no param) switching to prev mode
+            // keep compatibility w/ existing behavior of `mode` (no param)
+            // switching to prev mode
         } else if (strcmp(name, "prev") != 0) {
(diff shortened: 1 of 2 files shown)


artemsen:
Unfortunately, you've come at a bad time - I'm in the middle of a [major rework](https://github.com/artemsen/swayimg/tree/v5) of the project.
Changes merged into the master branch will be lost, so let's pause this PR and come back to it a little later after moving to the new version of the application.

artemsen:
The latest version of Swayimg allows to implement this feature in Lua, so we don't need to write any additional C++ code for it.

```lua
local last_mode = swayimg.get_mode()

local function set_last_mode()
  local prev = swayimg.get_mode()
  swayimg.set_mode(last_mode)
  last_mode = prev
end

swayimg.viewer.on_key("Backspace", function()
  set_last_mode()
end)
swayimg.gallery.on_key("Backspace", function()
  set_last_mode()
end)
swayimg.slideshow.on_key("Backspace", function()
  set_last_mode()
end)
```

Issue #401, opened 2026-03-10
Performance suggestion: use LuaJIT


litoj:
Since the switch to lua, swayimg has been much slower and less responsive. Though C++ may also be the cause of that, the default lua implementation is known to be slow. LuaJIT implements lua 5.1 in a much more efficient way (see https://github.com/LuaJIT/LuaJIT/issues/929 why not 5.4+ and why it would bring no benefit).

I noticed a 0.7s slowdown when opening swayimg when using my config, which is only a direct rewrite of my config from the previous toml format.

Could we switch to luaJIT? Would it help if I attempted to make a PR for it?

artemsen:
> Could we switch to luaJIT?

What do you mean? We already use [luajit](https://github.com/artemsen/swayimg/blob/ee1b80cdb48e9e36cd87c5850680a00ae03508bc/meson.build#L38).

litoj:
huh, so that must mean then, that the lua to cpp api methods are really slow. That's weird.

Sorry, I thought it uses PUCLua 5.4

litoj:
nevermind - the issue is with me using the on_initialized to conditionally load the adjecent files.

When I use `l.enable_adjecent(true)` it loads in about 0.3s, but with `swayimg.on_initialized(function() l.add(v.current_image().path:match'.+/') end)` it takes about 0.7s

I don't know if there is a way to mitigate that slowdown, but it seems weird that it would take so much longer even though the exact same amount of work is being done.

artemsen:
It's because manual adding is a complex function. It adds a file to ordered position, while initial loading sorts the list only once at the end of loading all entries.

litoj:
In that case it could be also changed to just add all the files first, and sort them before returning from the `add` call

artemsen:
By the way, `l.enable_adjecent(true)` in latest version works only at startup. Otherwise it breaks FS monitor and adds even removed entries.

PR #402, opened 2026-03-10
perf: sort files after all have been added

litoj:
fixes slowdown mentioned in #401 by first loading all files in a directory and sorting only after all loading is done.

Also avoids unnecessary on-return list copying.

### DIFF
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -547,7 +547,7 @@ void ImageList::add_entry(ImageEntryPtr& entry, const bool ordered)
 
 void ImageList::sort(bool locked)
 {
-    auto unlocked_sort = [&]() {
+    auto unlocked_sort = [this]() {
         if (order == Order::None) {
             // nothing to do
         } else if (order == Order::Random) {
@@ -572,6 +572,7 @@ void ImageList::sort(bool locked)
         reindex();
     } else {
         unlocked_sort();
+        reindex();
     }
 }
 


artemsen:
Even though your patch causes deadlock, I can't accept it.
The lua methods are designed to add files at runtime and they **must be sorted**.
I think you're trying to solve a very specific problem using the wrong methods. Use `--from-file` if you need to generate a list of files in advance.

litoj:
The deadlock is easily solveable. Of course the files must be sorted, I know that. But they don't need to be sorted with O(n²) - normal O(nlogn) sorting is just as good. This is not about changing behaviour. This is about making the behaviour be efficient.

artemsen:
Hmm.. You are right, sorry.
Please, fix deadlock and remove editor file.

src/imagelist.hpp
```diff
@@ -167,7 +167,12 @@ class ImageList {
     void add_entry(ImageEntryPtr& entry, const bool ordered);
 
     /**
-     * Sort image list.
+     * Sort image list without locking.
+     */
+    void unlocked_sort();
```
artemsen: Instead of new function, I would add `bool` argument.

litoj: The issue with keeping it just one function is that the lock is a local variable - meaning making the lock optional would require duplicating the entire code for the sorting, i.e.:
```cpp
// this would unlock immediately
if (lock) std::unique_lock lock(mutex);
... sort
if (lock) reindex();

// this would work, but the entire sort code would be there twice
if (lock) {
   ... lock, sort, reindex
} else ... sort
```

litoj: In order to do this without code duplication, I had to introduce an inner function (lambda). Please check if this solution is acceptable.

src/imagelist.cpp
```diff
@@ -547,27 +545,34 @@ void ImageList::add_entry(ImageEntryPtr& entry, const bool ordered)
...
+    if (locked) {
+        std::unique_lock lock(mutex);
+        unlocked_sort();
+        reindex();
     } else {
-        entries.sort([this](const ImageEntryPtr& l, const ImageEntryPtr& r) {
-            const bool cmp = compare_entries(*l, *r, order);
-            return reverse ? !cmp : cmp;
-        });
+        unlocked_sort();
```
artemsen, line 574: `reindex` should be called in any case

litoj: it is called later in the public `add()`, but you're right - for self-containment it should be here as well.

src/imagelist.cpp
litoj, line 521: @artemsen while in the hunt for performance, I'd also like to disable this when the list isn't being ordered and replace it with checking for same paths after one another after all files have been sorted.

i.e. changing processing order:
- from `x,a,x (removed),z -> a,x,z`
- to `x,a,x,z -> a,x,x (removed),z`

this would make just one linear pass at the end of loading files, instead of quadratic (checking the entire list for each file).

artemsen: Yeah, I know about this problem.
Let's leave it as is. I plan to replace `std::list` with `std::unorderd_map` to solve this problem.

litoj: But if you do that, how will you maintain the order? Are you sure this isn't a simpler solution?

artemsen:
Merged, thank you!
Special thanks for all the bug reports that helped me fix the errors =)
I'll wait a couple of days and if nothing else is found, I'll release a new version of the Swayimg.

litoj:
Thanks to you, too. Since I've switched to linux I couldn't find an image viewer that would suit my needs until I found this gem that you so relentlessly keep working on and improving.

I will try to figure out that problem of initial size that I have and make a PR if I figure anything out.

src/imagelist.cpp
artemsen: I haven't thought about it deeply yet, but I'll come up with something. Maybe it won't be a replacement, but a duplicate. Too many lookup operations, especially when modifying the list at runtime. In any case, this is an optimization for the next version.

litoj: I tried implementing my deduplication behaviour. It isn't really necessary at all.

On a list of 2135 images I save exactly 10% - from 1.2s to 1.08s.

```
add_entry()...
    if (ordered || order == Order::None || order == Order::Random)

sort()...
            std::filesystem::path last;
            auto it = entries.begin();
            auto end = entries.end();
            while (it != end) {
                if ((*it)->path == last) {
                    it = entries.erase(it);
                } else {
                    last = (*it)->path;
                    ++it;
                }
            }
```

Issue #403, opened 2026-03-10
[FR]: make config options accessible as variables


litoj:
**Is your feature request related to a problem? Please describe.**
We used to be able to cycle through certain options or toggle them. This behaviour was mostly gone even in the last few pre-lua version of swayimg. However, I think with this update we should bring it back.

Currently, if I want to make a keybind to toggle an option without a getter, I have to make a local variable, like this:
```lua
	local alias = false
	swi.enable_antialiasing(alias)
	amap('u', function()
		alias = not alias
		swi.enable_antialiasing(alias)
	end)
```

**Describe the solution you'd like**
Use a lua-first approach, making the C variables also appear as lua variables. Internally, this can be done with a metatable forwarding the I/O to the respective methods, whether still in lua, or already in C.

Neovim uses this approach well - `vim.bo`=buffer options, `vim.go` = global, `vim.wo` = window, etc.

This will have the benefit of enforcing a consistent naming, leading to less confusion when looking for which variables are available or a name of a variable that the user might want to set but doesn't know the precise name. It will also allow to be closer to the original config style and naming (though some variables changed namespaces).

**Describe alternatives you've considered**
Creating the metatable in userspace lua.
```lua
local meta = {
	__index = function(self, idx) return self._api[idx] or self._api['get_' .. idx]() end,
	__new_index = function(self, idx, val)
		(type(val) == 'boolean' and self._api['enable_' .. idx] or self._api['set_' .. idx])(val)
	end,
}
_G.o = setmetatable({}, {
	__index = function(self, idx)
		local val = swayimg[idx]
		if type(val) == 'function' then return val end
		if val == nil then return swayimg['get_'..idx]() end

		self[idx] = setmetatable({ _api = swayimg[idx] }, meta)
		return self[idx]
	end,
	__new_index = meta.__new_index
})

-- usage:
o.text.size = 20
o.antialiasing = true
```
This has the downside of having no lsp docs for the variables and it has a very limited usability since the methods don't have consistent naming with their type (enable(true/false) vs show()/hide(), get_xxx vs current_image(), etc.).

**Additional context**
I know this is a lot of work. But for the user experience and everyone who will be converting from old toml config to lua this will be a life-saver. Personally, I've spent 8 hours trying to get 150-lines-long config ported to lua (mostly key bindings), a signifficant part of which was trying to find the right method to match the old setting or looking for a getter for a variable to get back the old toggle functionality, which were often not available.

I can make a PR if you'd be willing to consider it :)

[...]

PR #404, opened 2026-03-10
feat(lua): api v2 with implementation of a conversion layer

litoj:
hacks in #403 

Changes:
- removed `set_meta` since now if the image metadata actually changes, the object gets updated
  - this enforces that no fake metadata is set that is not actually in the file
- moved `set_status` under `text.set_status` because that's where the `status_timer` is already and it is related
- made all hook callbacks removable (`on_image_change`,`on_signal`,`on_window_resize`)
- joined `on_key` and `on_mouse` bindings into a typical `map()` function
- added a user-friendly shell `exec` function to mimic the old one from `ini` config
- renamed `viewer.open` to `viewer.select` to be consistent with `gallery` and be more representative of the actual action (we're not launching xdg-open or sth else to open the image, we're just selecting a different image in the current window)
- `imagelist.mark` -> `imagelist.mark_current`
- renamed img attribute `mark` to `marked` because the mark is not a string but a toggle
- moved `xxx.current_image` to `imagelist.get_current` since the other getter is there too and renamed `viewer.current_image` to `viewer.get_loaded_image`
- joined text timer manipulation to set a bool or number and remove `hide/show()`
- everything volatile stays as methods, the rest got converted to variables, meaning all will need a getter, too
- I provided a fully functional conversion layer that overcomes the missing getters by caching the last-set value to be able to read it.

Note: I used AI to help with the conversion layer and the docs, but I checked then myself that nothing is missing (except the intentional `set_meta` removal) and methods work. But it is always possible I've overlooked sth.

### DIFF
--- a/extra/swayimg.lua
+++ b/extra/swayimg.lua
@@ -164,12 +164,12 @@ local mode_base = {}
 
 ---Map a keyboard event to an action.
 ---@param bind string|string[] 1 or more mouse or keyboard events to map - `Alt+s`, etc.
----@param cb string|fun() shellcmd to execute (% for current image) or callback function to run
+---@param cb fun() callback function to run
 function mode_base.on_key(bind, cb) end
 
 ---Map a mouse event to an action.
 ---@param bind string|string[] 1 or more mouse or keyboard events to map - `Ctrl+ScrollDown`, etc.
----@param cb string|fun() shellcmd to execute (% for current image) or callback function to run
+---@param cb fun() callback function to run
 function mode_base.on_mouse(bind, cb) end
 
 ---Remove all existing key/mouse/signal bindings.


extra/v1_to_v2_mapper.lua
```diff
@@ -0,0 +1,241 @@
+--------------------------------------------------------------------------------
```
artemsen: 1. We haven't had a release with Lua support yet, so technically it's v0->v1.
2. We don't need this file at all, I will bind properties directly in c++ code.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
-swayimg.viewer = {}
-
----Slideshow mode.
----@class slideshow
-swayimg.slideshow = {}
-
----Gallery mode.
----@class gallery
-swayimg.gallery = {}
+---@meta swi
```
artemsen: Please revert name to `swayimg`.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
----| "viewer" # Image viewer mode
+---| "viewer"    # Image viewer mode
 ---| "slideshow" # Slide show mode
----| "gallery" # Gallery mode
+---| "gallery"   # Gallery mode
 
----Exit from application.
----@param code? number Program exit code, `0` by default
-function swayimg.exit(code) end
+---@alias tuple {[1]:integer,[2]:integer}
```
artemsen: Unfortunately, the Lua bridge used in Swayimg does not allow the use of tuples, so such double values ​​are a table.
That is, it cannot be written in a script `local a,b = cal()`, it should be `local a,b = unpack(call())`.
So, please revert it to table.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
----Set window title.
----@param title string Window title text
-function swayimg.set_title(title) end
+---Main application class.
+---@class swi
+---@field title string Window title text
+---Mouse button for drag-and-drop to external apps.
+---Configurable only at startup.
+---@field drag_button string
+---@field antialiasing boolean Enable/disable antialiasing TODO: specify which algorithm is used?
```
artemsen: No, there is only one algorithm.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
----@param status string Status text to show
-function swayimg.set_status(status) end
+---Exit from application.
+---@param code? integer Program exit code, `0` by default
+function swi.exit(code) end
 
 ---Switch to specified mode.
 ---@param mode appmode_t Mode to activate
-function swayimg.set_mode(mode) end
+function swi.set_mode(mode) end
```
artemsen: Can be a variable too.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
 
 ---Get current mode.
 ---@return appmode_t # Currently active mode
-function swayimg.get_mode() end
+function swi.get_mode() end
 
 ---Get application window size.
----@return table # (width, height) tuple with window size in pixels
-function swayimg.get_window_size() end
+---@return tuple # (width, height) tuple with window size in pixels
```
artemsen: It is a table, not a real tuple.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
----@param width number Width of the window in pixels
----@param height number Height of the window in pixels
-function swayimg.set_window_size(width, height) end
+---@param width integer Width of the window in pixels
+---@param height integer Height of the window in pixels
+function swi.set_window_size(width, height) end
 
 ---Toggle full screen mode.
 ---@return boolean # True if full screen is enabled
 function swayimg.toggle_fullscreen() end
```
artemsen: Can be variable.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
 ---Set the initialization completion handler.
 ---Called after all subsystems have been initialized.
 ---@param fn function Initialization completion callback
-function swayimg.on_initialized(fn) end
+function swi.on_initialized(fn) end
+
+---Call an external shell command.
+---Use `%` to refer to the current image path.
+---@param cmd string
+function swi.exec(cmd) end
```
artemsen: One of the main goals when switching to Lua was to get rid of this feature, but you brought it back.
Lua has built-in support for this.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
+
+--------------------------------------------------------------------------------
+-- Image object
+--------------------------------------------------------------------------------
+
+---A single image entry from the image list. READONLY object
+---@class image
```
artemsen: It's not an image. It's entry in image list,  it may be a text file.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
+---@field index integer
```
artemsen: It's an index in the image list.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
+---@field reverse boolean Reverse the sort order
+---@field recursive boolean Recursive directory reading
+---@field adjacent boolean Open adjacent files from the same directory
+swi.imagelist = {}
 
 ---Get size of image list.
----@return number # Number of entries in the image list
-function swayimg.imagelist.size() end
+---@return integer # Number of entries in the image list
+function swi.imagelist.size() end
```
artemsen: Can be variable

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
 
 ---Get list of all entries in the image list.
----@return table[] # Array with all entries
-function swayimg.imagelist.get() end
+---@return image[] # Array with all entries
+function swi.imagelist.get() end
+
+---Get information about currently selected image.
+---@return image # Currently selected image entry
+function swi.imagelist.get_current() end
```
artemsen: Image list doesn't have "current image", it doesn't even have "current entry". Implementation of this function is mode specific.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
+
+---Mark or unmark the current image.
+---@param state? boolean Should the image be marked or not
+function swi.imagelist.mark_current(state) end
```
artemsen: Image list doesn't have "current image". I was too lazy and put it here, but technically it should be in app mode.

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
-function swayimg.text.set_font(name) end
-
----Set font size.
----@param size number Font size in pixels
-function swayimg.text.set_size(size) end
+---Text overlay layer.
+---@class text
+---Toggle for the entire text layer (regardless of active mode).
+---Specify a number (float) to use a timeout in seconds after which the entire text layer is hidden.
+---@field visible boolean|number
```
artemsen: This is ambiguous. Lets split it into `visible` and `timer`

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
+---Add a callback function called when a new image is selected.
+---@param fn fun():(boolean?) Handler, return true to detach
+function mode_base.on_image_change(fn) end
 
----Set a timeout after which the entire text layer will be hidden.
----@param seconds number Timeout in seconds
-function swayimg.text.set_timer(seconds) end
+---Add a callback function called when main window is resized.
+---@param fn fun():(boolean?) Handler, return true to detach
+function mode_base.on_window_resize(fn) end
```
artemsen: Window size is in the main namespace. May be it would be better to move this hook there too?

extra/swayimg.lua
```diff
@@ -1,196 +1,203 @@
...
-function swayimg.text.show() end
+---Remove all existing key/mouse/signal bindings.
+function mode_base.bind_reset() end
 
----Hide text layer and stop the timer.
-function swayimg.text.hide() end
+---Map an input event to an action.
+---@param bind string|string[] 1 or more mouse or keyboard events to map - `Ctrl+ScrollDown`, etc.
+---@param cb string|function shellcmd to execute (% for current image) or callback function to run
+function mode_base.map(bind, cb) end
```
artemsen: It s a bad idea to make a common binding for all inputs. Lets split it as it was before: on_signal/on_key/on_mouse.

extra/swayimg.lua
```diff
@@ -235,375 +242,116 @@ function swayimg.text.hide() end
 ---| "mirror" # Fill window with the mirrored current image and blur it
 ---| "auto" # Fill the window background in `extend` or `mirror` mode depending on the image aspect ratio
 
----Open next file at specified direction.
----@param dir vdir_t Next file direction
-function swayimg.viewer.open(dir) end
+---@class loaded_image: image
```
artemsen: Technically, there are 3 loaded images by default, so this name should be `displayed image` for viewer/slideshow and `selected` for gallery.

extra/swayimg.lua
```diff
@@ -235,375 +242,116 @@ function swayimg.text.hide() end
...
+
+---Go to the next file in the specified direction.
+---@param direction vdir_t Next file direction
+function swi.viewer.select(direction) end
 
 ---Get information about currently viewed image.
----@return table # Dictionary with image properties: path, size, meta data, etc
-function swayimg.viewer.current_image() end
+---@return loaded_image # Currently viewed image entry
+function swi.viewer.get_loaded_image() end
```
artemsen: Not "loaded", but "displayed"

extra/swayimg.lua
```diff
@@ -235,375 +242,116 @@ function swayimg.text.hide() end
...
+---@field mark_color integer Mark icon color in ARGB format
+---@field freemove boolean Free move mode TODO: needs a more detailed explanation
+---@field loop boolean Image list loop mode
+---@field preload_limit integer Number of images to preload in a separate thread
+---@field history_limit integer Number of previously viewed images to keep in cache
+swi.viewer = {}
+
+---Go to the next file in the specified direction.
+---@param direction vdir_t Next file direction
+function swi.viewer.select(direction) end
```
artemsen: Not "select", but "open". There is nothing to select.

artemsen:
Our intentions were unexpectedly dashed against the wall of LSP.
I haven't found a way to create a multi-line field description other than using everything on one line interspersed with `br` (which is very inconvenient).
This is a blocking issue because the documentation (CONFIG.md) is generated from the lua file.

extra/swayimg.lua
litoj: it is just an alias, it is a table. lua_ls automatically expands aliases so the user will only see the table definition anyway.

extra/swayimg.lua
litoj: I know there is only one, but before we had multiple, so we could write which one is used here. Otherwise there is no reason to add redundant documentation to the variable - it is obvious that antialiasing boolean toggles the antialiasing, we don't need a comment for that

extra/swayimg.lua
litoj: We could. Honestly, in the name of consistency, I'd move all the hooks there (except mappings = key&mouse). If the user wants to target just one mode, thay can `if swayimg.mode == 'mode'`-out the other modes

extra/swayimg.lua
litoj: the on_signal is still separate, but the user inputs would make sense to be together - it makes it easier for the user to map them or to change the mapping, don't you think?

Practically, I was just adjusting the api to be more like it is in vim.

extra/swayimg.lua
litoj: yes, but _open_ is most often associated with opening the file in a different application. Swayimg is already opened, it is just switching to a different image. Although I understand where you're coming from, when I was looking for a way to switch images in viewer mode, I was looking for `prev/next` and `select`. I'd have never thought of looking up `open`.

extra/swayimg.lua
artemsen: I mean we may not specify the name. Just remove "TODO:..."

litoj:
> Our intentions were unexpectedly dashed against the wall of LSP.
I haven't found a way to create a multi-line field description other than using everything on one line interspersed with br (which is very inconvenient).

Are you sure this doesn't work either?

I did this only because when I do text completion I do see all the lines as they should be:
obrazek

extra/swayimg.lua
artemsen: I think it would be better to have mouse and keyboard binding separately, this removes possible name coincidences in the future.

extra/swayimg.lua
litoj: I can change the name of the class to sth like `list_entry` if you think that's better.

extra/swayimg.lua
artemsen: May be `switch_image`?

extra/swayimg.lua
litoj: yes, that is obvious from the field name. Do we really need redundant comments?

extra/swayimg.lua
litoj: I originally had it that way but decided against it since it is read-only and gets changed by outside means (adding a path).

artemsen:
> Are you sure this doesn't work either?

Yes, `lua-language-server --doc` looses this info, only one field's string in descriotion.

extra/swayimg.lua
litoj: Will fix, thank you for the explanation

extra/swayimg.lua
artemsen: We have many write-only variable here, so... =)

extra/swayimg.lua
litoj: Than you cannot have it as variables at all, since show/hide disable the timer. This way it clearly says what the condition of visibility for the text layer is.

Edit:

It wouldn't be a problem in the translation layer anyway, since you'd most likely already have the __new_index first split the vars by incoming type and then use the correct method,
 sth like this (simplified):
```lua
__new_index = function(_, idx, val)
  if type(val) == 'boolean then
    if idx == ... then
      ...
    elseif idx == 'visible' then
      c_text:set_visible(val)
    else
      error('Invalid api write to: '..idx)
    end
  else
    if idx == ... then
      ...
    elseif idx == 'visible' then
      c_text:set_timer(val)
    else
      error('Invalid api write to: '..idx)
    end
  end
end
```

extra/swayimg.lua
artemsen: Yes, this will became a documentation (CONFIG.md).

extra/swayimg.lua
litoj: I was considering it, too. However, its state can get changed by the compositor - therefore it can be indirectly changed, so I thought we shouldn't.

If we do decide to make this a variable, then I'd also like to make the viewer scale and pos into
 ```lua
---@field image_scale fixed_scale_t|number
---Position of the image relative to the position of the window.
---Example: ←↑ corner of the image is outside the window -> `x,y<0`
---@field image_pos fixed_position_t|{[1]:integer,[2]:integer}
```
Which I'd honestly prefer, but it would have to be both to keep at least some reason to it. I don't know if we should do the same for window_size at that point, though.

File: extra/swayimg.lua (hunk 16 above)
Comment (litoj) [reply extra/swayimg.lua]: I don't understand what name coincidence you could possibly mean. The binding style is identical for both, just the name of the key is different. And it will always be different, because internally the system handes mouse just as the same input as a keyboard, only with a much higher rate of event flow.

Should I really split it anyway?

File: extra/swayimg.lua (hunk 11 above)
Comment (litoj) [reply extra/swayimg.lua]: well, only in the sense that we cannot currently read them, but there is no reason why we shouldn't be able to.

With the exception of the `image_entry` table all variables are currently both read _and_ write.

**Comment by litoj** [comment]
I've resolved all the conversations where I made a change similar to- or exactly what was requested. All conversations that I think still deserve more conversation before a final decission are open.

`on_window_resize,on_image_change,on_signal` are now all in the top namespace (`swayimg`) and the user can filter out the mode they want to work with using `swayimg.mode`

I've also decided to make the coordinate/resolution getter methods be more like the setter methods, meaning no table is used and we just return multiple values.

**Comment by litoj** [comment]
> Yes, lua-language-server --doc looses this info, only one field's string in descriotion.

It did the full description in my case - I ran just the `lua-language-server --doc extra/swayimg.lua` and the `doc.md` as well as the `doc.json` contain full descriptions:
obrazek

File: extra/swayimg.lua (hunk 7 above)
Comment (artemsen) [reply extra/swayimg.lua]: > state can get changed by the compositor

We can get current state.

> make the viewer scale and pos into

Lets do it.

File: extra/swayimg.lua (hunk 14 above)
Comment (artemsen) [reply extra/swayimg.lua]: > Than you cannot have it as variables at all, since show/hide disable the timer. 

That why I suggest to split it. Timer is timer, visible is a current state.

**Comment by artemsen** [comment]
> It did the full description in my case - I ran just the `lua-language-server --doc extra/swayimg.lua` and the `doc.md` as well as the `doc.json` contain full descriptions

Wow! It work! I did something wrong in my experiment. Sorry.

File: extra/swayimg.lua (hunk 16 above)
Comment (artemsen) [reply extra/swayimg.lua]: Yes please.

**Comment by artemsen** [comment]
> `on_signal` are now all in the top namespace

No, i should be placed separately for viewr/gallery/slideshow. It like a keybindings, echh app mode has own scheme.

**Comment by litoj** [comment]
So should I revert just `on_signal`, or also `on_image_change`?

File: extra/swayimg.lua (hunk 7 above)
Comment (litoj) [reply extra/swayimg.lua]: > We can get current state.

I know, but I meant that the value can change even without the user explicitly setting it here.

> Lets do it.

Alright, I'm on it

Comment (litoj) [reply extra/swayimg.lua]: @artemsen What is your opinion on how we should treat the `window_size`?
It can change whenever the user moves the window or enables fullscreen etc.

Should it be a variable as well?

**Comment by artemsen** [comment]
> So should I revert just `on_signal`, or also `on_image_change`?

Both.
Also, I didn't understand the idea with boolean in callback
`@param fn fun():(boolean?) Handler, return true to detach`

File: extra/swayimg.lua [hunk 20]
```diff
@@ -1,122 +1,115 @@
...
-function swayimg.get_mode() end
+---Exit from application.
+---@param code? integer Program exit code, `0` by default
+function swayimg.exit(code) end
 
 ---Get application window size.
----@return table # (width, height) tuple with window size in pixels
+---@return integer width in pixels
+---@return integer height in pixels
 function swayimg.get_window_size() end
```
artemsen, line 39: It not possible to return tuple, only one object on stack. Table in this case.

extra/swayimg.lua
artemsen: No, it should be a function. It's not convenient to operate with table variables.

extra/swayimg.lua
litoj: > That why I suggest to split it. Timer is timer, visible is a current state.

If we make them into two variables, however, how would you solve the situation of `visible=false`, but `timer=3` and you want to turn off the text layer completely (permanently)?

if `timer=0`, then the visibility would be determined by the `visible` flag (I'm talking about when the user sets the values).

However, `timer>0` would mean that the `visible` flag is useless for writing, because it will get overwritten on the next image change. Therefore now the controlling flag of menu visibility is `timer`.

This means the meaning and purpose of the variables changes based on one another, which is exactly what should not be happening with config variables - each one should have their one purpose that stays the same no matter the conditions of the other variables.

But I do understand why you'd split it, too. In the current mockup, there'd be no way to find out if the text is currently visible (just like with the status message). For which I'd like to suggest a compromise. Sth like this:

```lua
---@class swayimg.text
---@field enabled boolean|number

function swayimg.text.is_visible() end
```

What do you think?

extra/swayimg.lua
litoj: You can do it in lua, so it is certainly possible in C as well. I've done it myself, you just call the same method multiple times and return the number of pushed values.

see <http://lua-users.org/lists/lua-l/2007-02/msg00016.html>

litoj: > Also, I didn't understand the idea with boolean in callback
@param fn fun():(boolean?) Handler, return true to detach

This is to be able to detach the handler - when the callback function returns `true` it means it doesn't want to listen to the event anymore

my implementation of this is like this:
```lua
local function gen_unhookable(fn_name)
	local hooks = {}
	local initiated
	return function(fn)
		if not initiated then
			local function master_hook()
				local i = #hooks
				while i > 0 do
					if hooks[i]() then table.remove(hooks, i) end
					i = i - 1
				end
			end
			swi.gallery[fn_name](master_hook)
			swi.viewer[fn_name](master_hook)
			swi.slideshow[fn_name](master_hook)
			initiated = true
		end
		hooks[#hooks + 1] = fn
	end
end
```

extra/swayimg.lua
artemsen: visible=true: force show, timer off
visible=false: force hide, timer off
timer=3: enable timer again
visible==true && timer=3 is ok, timer don't decreased in runtime, it is not the elapsed or rest time, it's timer setup.

extra/swayimg.lua
artemsen: > certainly possible in C as well

As I said earlier, it is a Lua bridge limitation. I don't want to rewrite this lib now.

artemsen: > This is to be able to detach the handler

Let's keep things simple. We don't support one-time hooks yet, at least not in the first version. Please remove it.

extra/swayimg.lua
litoj: I think it would make sense to be able to measure the size like in normal lua by implementing the `__len` metamethod.

artemsen: I don't think that `__len` is a normal name for a public function.

extra/swayimg.lua
litoj: Well, then we cannot make the position be a variable either. At which point we shouldn't be making the scale a variable either because the scale and position have the exact same set of operations - fixed and absolute values.

I am not against either a method or table, but it has to be all 4 of them, since they share the same environment conditions - can change based on outside conditions (compositor event, moving to next image that resets the scaling etc.)

extra/swayimg.lua
litoj: it's the metamethod that implements the `#` operator, not a public function. (just like `__index` etc.)

extra/swayimg.lua
litoj: That is true, but the variables influence each other, which is not something a variable should be able to do. When you set a variable you expect to set just the variable, not change multiple variables. That's what functions are for - behaviour more complex than a primmitive assignment.

extra/swayimg.lua
litoj: > > certainly possible in C as well
> 
> As I said earlier, it is a Lua bridge limitation. I don't want to rewrite this lib now.

I still don't understand. Is it possible in C but not in C++? Because clearly they do support it.

From https://github.com/vinniefalco/LuaBridge/issues/238#issuecomment-690948658 I understand it would be a problem if you were using that method in other than pure lua context. If that is the case, than it would be annoying to have to do a separate method indeed.

extra/swayimg.lua
artemsen: So, do you want to replace `size` to `__len from swayimg.imagelist` ?

extra/swayimg.lua
artemsen: I don't have any ideas how to make better. If we don't disable timer on `visible=true`, the text layer will hide again. Asking a user to call timer=0 and then visible=true is not a good solution.

extra/swayimg.lua
litoj: yes, so that the user can do `#swayimg.imagelist`.

or we can keep it a method, either way works.

I just don't want to create a readonly variable when all the other variables of that object are R+W.

I'm not sure, though, what a metatable support looks like with luabridge, because the `__len` method isn't in the same object as the main table, but in a special metatable, like this:
```lua
imagelist = setmetatable(
  { order =xxx, get = function() end, ... },
  { __index = function()end, ...,
    __len = function(self) return #self.get() end
  }
)
```

extra/swayimg.lua
artemsen: It is not the lib we use =)
https://github.com/kunitoki/LuaBridge3/discussions/74

extra/swayimg.lua
artemsen: It looks ugly =) Let's leave normal `size`.

extra/swayimg.lua
litoj: That is exactly why I want to put it in a single lua variable that can just have different values, but all of which clearly and unanymously declare the condition under which the text layer can be visible.

And for the immediate state I've added the `is_visible()` method to be able to query it.

I've renamed the variable to `allowed` which goes more in line with what it's actual purpose is - it sets the condition for when it can render at all. It doesn't say if it is currently rendered.

extra/swayimg.lua
litoj: can I keep the size a function then?

extra/swayimg.lua
artemsen: Toooo complex. Even I completely confused with timers, flags, and now "allowed".
Please, let's leave two fields: timer (number) and visible (boolean).

extra/swayimg.lua
artemsen: Yep!

artemsen:
Let's take a break.
I have to switch to another project, but I'll get back to Lua on the weekend.

extra/swayimg.lua
litoj: but that is literally what you agreed isn't a good solution if we treat variables as just arbitraty assignments with no magic behind:

> If we don't disable timer on visible=true, the text layer will hide again. Asking a user to call timer=0 and then visible=true is not a good solution.

It isn't ergonomic because the two variables both serve the same end goal through different means.

The behaviour that defines how the text layer is displayed is just one subject, therefore it makes sense to keep it as just one variable. The question is only how to call it and what the best self-explanatory values are.

Some ideas for improvement:
```lua
-- described numerically
-- number meaning is clear, but edge cases 
-- require a quick (though logical) thought
-- -1=always on, 0=always off, 5=time to off
timeout = -1|0|5

-- more descriptive name
auto_hide_timeout = -1|0|5
-- verbal description of the time period
-- varname describes the number
-- strings carry a temporal meaning to fit in
auto_hide_timeout = 'never'|'no-display'|5

-- include 0 since its logical conclusion is 
-- that it never shows the text layer
hide_timeout = 'never'|0|5

--without inverted logical meaning
-- (not when to hide, but how long to keep displayed)
text_timer    = 1/0 (inf)   |0|5
text_duration = 'forever'   |0|5
overlay_timer = 'infinite'  |0|5
display_timer = 'permanent' |0|5
```

If none of these seem fitting, then I'd also be for going back.
- Either to methods as they were before,
- or better, give `enabled` the top priority and make timer effective only with `enabled=true`, with `timer=0` meaning it is always visible.

extra/swayimg.lua
```diff
@@ -160,12 +164,12 @@ local mode_base = {}
 
 ---Map a keyboard event to an action.
 ---@param bind string|string[] 1 or more mouse or keyboard events to map - `Alt+s`, etc.
----@param cb string|function shellcmd to execute (% for current image) or callback function to run
+---@param cb string|fun() shellcmd to execute (% for current image) or callback function to run
```
artemsen: There is no `shellcmd to execute (% for current image)`, just a callback

litoj: I know, that's the feature I wanted to add, because half of the mappings one will have will be shell commands, so they could just provide the string of the shell command to map.

But I understand you wanted to force the user to execute the commands themselves.

Perhaps in some future release we may see it again? =`)

Issue #405, opened 2026-03-11
letter bindings aren't repeatable with Shift


litoj:
binding `i` to zoom in works as expected. Meaning holding it down keeps zooming in repeatedly. However, when a binding of a simple letter (a-z) contains the `Shift` modifier, then the binding doesn't work with the auto-repeat. It must be pressed physically multiple times. Holding it down prints the message that such combination is not bound (`Unhandled key: Shift+i`), regardless of any additional modifiers (`Ctrl/Alt/Super` etc.).

My wild guess would be that swayimg receives the event lowercase first and then uppercase on the repeated presses based on the fact that the non-letter keys must always be mapped in their shifter version when mapped with the shift modifier (on US `minus`, but with shift it is `Shift+underscore`)

RgeditV1:
working on it

RgeditV1:
@litoj the problem is that holding Shift converts the key to Caps, and 'i' is not the same as 'I'. That’s why it fails; when you hold Shift + i, it eventually registers as Shift + I, which isn't a valid keybind.

ill try to fix it

RgeditV1:
i think i got it
ill do pull request

RgeditV1:
@litoj done, but its depend of your keyboard layout, currently its working with english - international

artemsen:
Fixed in eff4a9a66e7314980bc5451672dac3652ef87376

PR #406, opened 2026-03-12
Fix Shift KEYMOD BUG

RgeditV1:
fix bugs from issue #405

- [x] now Shift won't capitalize the keys anymore.


litoj:
Do not add individual keys. That's not a portable solution. The lowercase is all that's necessary.

Thank you for looking into it.

Issue #407, opened 2026-03-13
incorrect alpha sorting


litoj:
Using `swayimg.imagelist.set_order('alpha')` I expect files to be ordered like they are in the output of all file tools I know - by the integer value of the characters. That means the order should follow the ASCII table.

However, swayimg `alpha` sorting gives precedence to letters before numbers, even though they are the other way around in the ascii table. This often makes the viewing experience confusing when opening a directory from a file manager.

RgeditV1:
the program should use ascii table for sort?

litoj:
No, the ascii table is just a good reference for what the correct order of the characters is.

RgeditV1:
> No, the ascii table is just a good reference for what the correct order of the characters is.

i think i got it

RgeditV1:
@litoj this is my output using the current alpha sort
```bash
001 /tmp/swayimg-sorttest/10.png
002 /tmp/swayimg-sorttest/1.png
003 /tmp/swayimg-sorttest/20.png
004 /tmp/swayimg-sorttest/2.png
005 /tmp/swayimg-sorttest/a.png
006 /tmp/swayimg-sorttest/b.png
007 /tmp/swayimg-sorttest/c.png
001 /tmp/swayimg-sorttest/1.png
002 /tmp/swayimg-sorttest/10.png
003 /tmp/swayimg-sorttest/2.png
004 /tmp/swayimg-sorttest/20.png
005 /tmp/swayimg-sorttest/a.png
006 /tmp/swayimg-sorttest/b.png
007 /tmp/swayimg-sorttest/c.png
--- /tmp/order-es_DO.txt        2026-03-13 10:54:35.809751130 -0400
+++ /tmp/order-C.txt    2026-03-13 10:54:35.862344226 -0400
@@ -1,7 +1,7 @@
-001 /tmp/swayimg-sorttest/10.png
-002 /tmp/swayimg-sorttest/1.png
-003 /tmp/swayimg-sorttest/20.png
-004 /tmp/swayimg-sorttest/2.png
+001 /tmp/swayimg-sorttest/1.png
+002 /tmp/swayimg-sorttest/10.png                                       +003 /tmp/swayimg-sorttest/2.png
+004 /tmp/swayimg-sorttest/20.png
 005 /tmp/swayimg-sorttest/a.png
 006 /tmp/swayimg-sorttest/b.png
 007 /tmp/swayimg-sorttest/c.png
```
ill try to use ascii as reference to implement

[...]

PR #408, opened 2026-03-13
fix alpha sort

RgeditV1:
fix for issue #407
new function compare_alpha that will work only if (!numeric)

Signed-off-by: Rgeditv1 <rgeditv1@protonmail.com>

### DIFF
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -15,7 +15,28 @@
 #include <string>
 
 /**
- * Comparison of two localized strings.
+ * Comparison of two strings using ASCII table order (byte-wise).
+ * @param l,r strings to compare
+ * @return compare result
+ */
+static int compare_alpha(const std::string& l, const std::string& r)
+{
+    const size_t n = std::min(l.size(), r.size());
+    for (size_t i = 0; i < n; ++i) {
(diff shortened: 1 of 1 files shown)


src/imagelist.cpp
```diff
@@ -62,11 +83,9 @@ static int compare_strings(const std::string& l, const std::string& r,
             ls.erase(0, ldend);
             rs.erase(0, rdend);
         }
+        return cmp;
     }
-
-    return cmp;
 }
-
```
litoj: Please add this empty line back. It separates the two methods.

Other than that, it looks great. The comparison could still be optimized a little further, but it certainly does the comparison correctly now.

Thanks!

RgeditV1: oky brother, hope this get merge

artemsen:
Sorry, I rushed the merge.
I reverted this commit because it breaks the sorting functionality.
When using international characters and UTF, the ASCII approach does not work.
For example, in Cyrillic, the alphabet sequence `е-ё-ж` doesn't correspond to the UTF sequence, and comparing character codes in this case is impossible. This is why complex constructs with std::locale and std::collate are used here, rather than just comparing chars.

Issue #409, opened 2026-03-14
Window background color is ignored in slideshow mode


mallniya:
**Describe the bug**
Window background color (`viewer.window` in config) works fine in viewer mode, but ignored in slideshow mode: it's always blurred current image on the background. Switching mode to viewer on the fly does activate color from `viewer.window`.

**To Reproduce**
Steps to reproduce the behavior, for example:
1. Set `viewer.window` to arbitrary color
2. Start `swayimg -l photo.jpg`
3. See blurred background image, NOT expected selected color
4. Press `Enter` to enter viewer mode
5. See selected color in background

**Version:**
4.6

RgeditV1:
ok, i think is for this line

```bash
Slideshow::Slideshow()
{
    // default settings

    set_history_limit(0);
    //set_window_background(Background::Auto);  //Here
    default_scale = Scale::FitWindow;

    text_scheme[static_cast<size_t>(Text::TopLeft)] = { "{name}" };

    bind_input(InputKeyboard { XKB_KEY_s, KEYMOD_NONE }, []() {
        Application::self().set_mode(Application::Mode::Viewer);
    });
}
```

its force to use background blur and not the custom color

RgeditV1:
@mallniya
Do you want swayimg to prioritize your init.lua script for the viewer/slideshow background?
 some like this?

of course you can set another color and will work
by default use black color, but in doing some changes

<img width="935" height="533" alt="Image" src="https://github.com/user-attachments/assets/c0ba51ed-7138-4e0c-a8a5-969f9cc25a2b" />
<img width="681" height="714" alt="Image" src="https://github.com/user-attachments/assets/a8e9b0b5-df1f-4052-933c-1072b8ec890d" />

artemsen:
> Window background color (`viewer.window` in config) works fine in viewer mode, but ignored in slideshow mode

Slideshow settings are stored in the `slideshow` section. Use `slideshow.window`.

PR #411, opened 2026-03-15
fix: remove hardcoded Background::Auto override in Slideshow constructor

RgeditV1:
issue #409  

The Slideshow constructor was unconditionally calling
set_window_background(Background::Auto), which forced the slideshow
mode to always display a blurred/mirrored background regardless of
user configuration. Since Slideshow and Viewer are separate singletons,
any window background color set via swayimg.viewer.set_window_background()
only affected the Viewer instance, while the Slideshow instance always
reverted to Background::Auto.
 
By removing this hardcoded override, the Slideshow now inherits the
default window background from its Viewer base class (solid black),
and respects user-configured background settings applied through
swayimg.slideshow.set_window_background().

The CI build failed due to a clang/glib compatibility issue in src/formats/svg.cpp (C2y extensions error), which is unrelated to my changes in src/slideshow.cpp, i just delete a line

### DIFF
--- a/src/slideshow.cpp
+++ b/src/slideshow.cpp
@@ -17,7 +17,6 @@ Slideshow::Slideshow()
     // default settings
 
     set_history_limit(0);
-    set_window_background(Background::Auto);
     default_scale = Scale::FitWindow;
 
     text_scheme[static_cast<size_t>(Text::TopLeft)] = { "{name}" };


artemsen:
1. The bug you're trying to fix is ​​related to an old version 4.6.
2. This isn't a bug, but a misconfiguration.
3. Everything works as designed, slideshow has own settings that slightly differs from viewer.

artemsen:
> The CI build failed due to a clang/glib compatibility issue in src/formats/svg.cpp

Yeah, librsvg problem, I'll fix it.

Issue #414, opened 2026-03-15
text layer doesn't update after adding files


litoj:
currently, when files get added the info line that shows the total files still stays at its original value until the image is changed.

test config:
```lua
swayimg.text.set_timeout(0)
swayimg.imagelist.enable_adjacent(false)
swayimg.viewer.set_text('topright',{'{list.total}'})
swayimg.on_initialized(function()
  swayimg.imagelist.add(swayimg.viewer.get_image().path:match'.+/')
end)
```

1. Open swayimg with a particular image (not folder) in a folder where multiple images are present.
2. observe the topright corner still says 1

RgeditV1:
what u think about add an "Auto-discovery" if you force to use adjacent to false

RgeditV1:
i got it

PR #415, opened 2026-03-16
fix: refresh list.total count when images are added

RgeditV1:
# From issue #414

## The Problem
When opening a single image with `adjacent` file scanning disabled (the default or via `swayimg.imagelist.enable_adjacent(false)`), the `ImageList` is initialized with a size of **1**. 

If a Lua script subsequently adds more images to the list using `swayimg.imagelist.add()`, the internal `ImageList` grows correctly, but the **Viewer's UI state remains stale**. The text overlay continues to display `1/1` because the Viewer has not been notified to re-index its current position within the now-larger list.

### The Solution
This PR ensures that when a `FileCreate` event is processed (which is triggered by Lua's `add` function), the Viewer performs a "re-sync". 

By calling `open_file` with the currently displayed image entry, we force the Viewer to:
1. Re-scan the updated `ImageList`.
2. Locate the current image's new index.
3. Updat

https://github.com/user-attachments/assets/161b9b02-d63a-4640-85fd-50dbc3de41a6

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -460,6 +460,10 @@ void Application::handle_event(const AppEvent::FileCreate& event)
     } else {
         entries = il.add(event.path);
     }
+    if (active_mode == Mode::Viewer) {
+        const ImageEntryPtr current = Viewer::self().current_entry();
+        Viewer::self().open_file(ImageList::Dir::First, current);
+    }
     for (auto& it : entries) {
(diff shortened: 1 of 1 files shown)


litoj:
We need to also refresh list.index - sorting could alter its position.

RgeditV1:
> We need to also refresh list.index - sorting could alter its position.

im lost with that :C

Issue #418, opened 2026-03-16
Swayimg won't open a file


litoj:
after one of the recent commits, swayimg won't start if it was launched with a file (not folder) argument.

No config required - happens always.

Errors when testing
```log
0 $ swayimg Pictures/Darktable/2026-02-23_100507.jxl
free(): unaligned chunk detected in tcache 2
fish: Job 1, 'swayimg Pictures/Darktable/2026…' terminated by signal SIGABRT (Abort)
134 $ swayimg Pictures/Darktable/2026-02-23_100507.jxl
fish: Job 1, 'swayimg Pictures/Darktable/2026…' terminated by signal SIGSEGV (Address boundary error)
139 $ swayimg Pictures/Darktable/IG_2025-08-24_133452_02.jpg 
malloc(): unaligned tcache chunk detected
fish: Job 1, 'swayimg Pictures/Darktable/IG_2…' terminated by signal SIGABRT (Abort)
134 $ swayimg Pictures/Darktable/IG_2025-08-24_133452_02.jpg
fish: Job 1, 'swayimg Pictures/Darktable/IG_2…' terminated by signal SIGSEGV (Address boundary error)
```

artemsen:
Do you use latest master version?
Looks like it was fixed in 8ba60569dd (I had to revert part of your commit =))

litoj:
that wasn't my commit, but yes, the attempt to fix the list update is the commit that causes the error

RgeditV1:
same here

litoj:
sorry, I read it incorrectly. The commit that causes the issue is e3e0ce8c3ca11919d02c4c9b72fc427fef880bc4

I tried compiling 6d4daaffd06b72ea67294d58149875e69ed9faab and that one still works

RgeditV1:
afte  the update i got this

```bash
➜ swayimg -v
--- [DEBUG] Iniciando test de configuracion ---
fish: Job 1, 'swayimg -v' terminated by signal SIGSEGV (Address boundary error)
```

litoj:
I assume this is due to sth like the text being set and updated before the text layer is initialized at all.

BTW that commit also just duplicates almost the same code three times, certainly not something that will help maintainability. It should be refactored to a standalone method

RgeditV1:
> I assume this is due to sth like the text being set and updated before the text layer is initialized at all.

what about use my commit and directly edit it

litoj:
> Do you use latest master version?
[...]

PR #419, opened 2026-03-16
fix duplicated code from viewer and aplication

RgeditV1:
from issue ##418

### DIFF
--- a/src/viewer.cpp
+++ b/src/viewer.cpp
@@ -786,14 +786,6 @@ void Viewer::preloader_start()
             }
         }
 
-        // update image list text info
-        Text& text = Text::self();
-        text.set_field(Text::FIELD_LIST_INDEX,
-                       std::to_string(current_image->entry->index));
-        text.set_field(Text::FIELD_LIST_TOTAL, std::to_string(il.size()));
-        text.update();
-        Application::redraw();
-
         image_pool.stop = true;
     });
 }


litoj:
there is still the second duplication in `application.cpp`, are you sure that cannot cause an issue ever?

Also, the same update is necessary in gallery mode, because you can see the item index and list size, too. So this fix doesn't work

litoj:
The problem is, that these aren't fixes in the meaning that they fix the underlying issue that causes them, they are just workarounds. The real fix would be to ensure less code coupling between all the classes. Right now if I knew about a tool that could print the file dependency graph, I'm pretty sure it would be close to an n-n web of everything touching everything, which is what makes those issues much harder to debug.

Of course with all that is going on, doing sth about it is really difficult, but we need a different structure to allow both-ways updates for each file, and then each file will register event listeners to each other file for the parts that they need. Practically reversing the dependencies, so that the base code won't have any dependencies and the other code can hook/listen into its changes.

In particular, we need an imagelist hook for list updates that the other files will hook into. And the imagelist should be the singular source of truth about what updates happened to the files - not the `application.cpp` that currently touches other files to insert fixes in all places where it updates them, but rather those files/classes that get affected, should register hooks with imagelist so that they can adapt accordingly.

RgeditV1:
> The problem is, that these aren't fixes in the meaning that they fix the underlying issue that causes them, they are just workarounds. The real fix would be to ensure less code coupling between all the classes. Right now if I knew about a tool that could print the file dependency graph, I'm pretty sure it would be close to an n-n web of everything touching everything, which is what makes those issues much harder to debug.

i can delete the if, but that all i can do, i need to investigate it more :c

src/viewer.cpp
```diff
@@ -786,14 +786,6 @@ void Viewer::preloader_start()
             }
         }
 
-        // update image list text info
```
artemsen, line 789: It is not duplication. On line 778 the image entry may be removed, so we should update the counter.

litoj:
I'm working on it. It needs a solid structure rewrite to allow this to happen without more coupling.

artemsen:
Is this patch really fix #418?

src/viewer.cpp
RgeditV1: could not be, but when i removed those lines, swayimg work again

RgeditV1:
> Is this patch really fix #418?

as @litoj said
it an alt of it

but it think need a refactor to work properly

currently im getting index fine

PR #422, opened 2026-03-17
Fix text alignment + popup

### LINKED ISSUE #399: font line height signifficantly larger than on 4.7.1
With the current swayimg C++ rewrite, font renders differently.

Previously I used font size 15, now I use 20 to get approximately the same size. This suggests now the size means px, whereas before it meant pt (dpi-agnostic unit). Just an observation, no issue here.

However, the line height is noticeably larger. The info lines now appear more separated making it more difficult to read when stretched across the image (+ it also obstructs the image more).

Image

### LINKED ISSUE #413: set_status doesn't work until first render
Lua errors and custom messages don't get printed on the screen during startup if they occured before the initial render of the window.

test config:
```lua
swayimg.text.set_status('test')
error('errortest')
```
the test message won't appear at all, whereas the error will get printed only to the terminal.

litoj:
Resolves #413
Resolves #399

Previous height adjustment was fixed because there was no math to get the correct height and center the text around it.
Now the height / text density is freely adjustable because the scaling works only with the base text size and therefore can be coupled with the horizontal padding to keep the text centered at all times. (not the best explanation, I think the one I've put in the code is better)

And since it is no longer a hack but an understood and computed behaviour I thought we might as well add the option to set the line spacing to allow both the old spacing and the latest one, defaulting to the old one to keep the experience for pre-lua users unchanged.

### DIFF
--- a/src/font.cpp
+++ b/src/font.cpp
@@ -8,7 +8,6 @@
 
 #include <fontconfig/fontconfig.h>
 
-#include <iostream>
 #include <memory>
 #include <utility>
 


src/font.cpp
```diff
@@ -156,10 +161,19 @@ Pixmap Font::render(const std::string& text) const
...
-    const size_t height = height_base + height_base / 3; // dirty hack
+    const size_t height = height_base * this->height_factor;
+
+    // Chars use a 1/6 extra space above an bellow the normal line (ygpq, ČÓŮ).
+    // We scale without it to be able to set the horizontal padding to
+    // the exact same value by offsetting it after scaling as well.
 
-    // horizontal padding
-    const size_t hpadding = height_base / 3;
+    // Using /2 to keep the text centered
```
artemsen, line 170: What are these constants based on?

src/text.cpp
```diff
@@ -33,13 +33,12 @@ Text::Text()
     foreground = { 0xff, 0xcc, 0xcc, 0xcc };
     background = { 0x00, 0x00, 0x00, 0x00 };
     shadow = { 0xd0, 0x00, 0x00, 0x00 };
+
+    font.load("monospace");
```
artemsen, line 37: Now font will be loaded event you just run `swayimg --version`.

litoj: Ah, I see. But does it matter? It doesn't sound like such a problem compared to the user getting confused by missing an error in his config, don't you agree?

I can try to find a better way, but with how connected everything is currently, I feel like I'd end up having to initialize it separately from a different file just before the config gets loaded, adding another (random) connection between the source files.

artemsen: > But does it matter?

Yes. Now it is loaded even if it not used. Even if user set another font in config. Let me fix it myself.

src/font.cpp
litoj: To be honest, just visual observations of multiplehours of pixel peeping. Right now a height factor of 4/3 will produce identical look to your original code.

I tested different values and noticed that at 1+1/6 lineheight the bottom still gets fully rendered. And since we can also cut 1/3 of space just from the top, it means I can cut 1/6 from both sides while keeping the text perfectly centered in the line and keep it centered even when scaling the lineheight beyond that. The /2 is there to shift it only by half of the factor, so that it stays centered.

And then, for fonts that use more vertical space, we can change the line height in the user config without forcing all users to have super spaced-out lines. That is the best out of both worlds.

src/text.cpp
litoj: Well, you cannot know what font the user wants if it is supposed to be loaded before user config. When their option gets processed, thd font will still get updated same as before.

Or maybe I misunderstood your usecase.

src/font.cpp
artemsen: So, you've replaced a one-line hack with a multi-line one.
In addition, it is now wrapped in a Lua script to generate new problems.
As I [said before](https://github.com/artemsen/swayimg/issues/399#issuecomment-4029417032), I don't want to make this hack public.
Please, remove it.

litoj: When it wasn't a hack in 4.7 and looked exactly the same as now with lineheight=1, why is it a hack now?

artemsen: Because this hack in 4.7 causes a crash when using some fonts.

src/text.cpp
artemsen: Lazy loading. Load a font only if it's needed but hasn't been initialized yet.

src/font.cpp
litoj: Yes, and the fix until now was called a hack why? Because of using a constant to adjust it? The problem will be always the same for any user with any font - fonts are just different therefore they require different settings to make them look the best. What you're saying is that the user shouldn't be able to change anything because it might crash with their super tall font, so we solve it by using a hack, instead of making it the user's responsibility to adjust the font height to fit his font correctly.

If you want to avoid the crash, I can implement a simple fix - just check for `y<0` and print out a message instead, but almost every GUI app I know (IDEs, gtk, etc.) lets you adjust it, so why should that improvement be forbidden to the users here?

It's not like the font cannot cause a crash ever - try using `LANG=C` and putting a `` star emoji at the end of a line (I used it to mark the units of Rating). It will also crash, but due to the width. The problem there is with how weird locales clearly are, because using `LANG=en_US.UTF-8` or any other lang works just fine.

That is why it should be left to the user, so they can adjust it if they have special conditions. Not make the special conditions mitigation be the standard for everyone. That would be like using only sign language to talk, because some people might not hear, but you'd still have the occasional blind individuals, and what then? What makes it a hack is that it isn't configurable.

artemsen: > Yes, and the fix until now was called a hack why? Because of using a constant to adjust it?

Yes, because of magic constant.

> I can implement a simple fix

[Already implemented](https://github.com/artemsen/swayimg/blob/aea6b7a61a39a7a4d31a658133eac65ad5f8d6b3/src/font.cpp#L180). The problem is that this is a hack too. This is symptom relief, not disease treatment.

> almost every GUI app I know (IDEs, gtk, etc.) lets you adjust it

They adjust mostly line spacing, not line height.
You are welcome to create PR that removes magic constants and render font as it should be rendered. Then we can add configurable spacing.

> so why should that improvement be forbidden to the users here?

Because adding height setup that works only with your fonts set doesn't resolve the problem, but adds new.
Just for example, tomorrow someone will create an issue that when setting the height, there is empty space at the top and cut off letters at the bottom. And will have to add "one more workaround". Sorry, but no.

litoj: Please, let me reword my reasons behind my intentions once more, now in the purely technical and causal manner.

The 1/6 differences are there because that is the height of every font  beyond the base text line (2/3 for the base line and 1/3 split for the top and bottom of the text) and I added it to the horizontal padding, so that it appears that the text is always on a 45° diagonal. Your so-called "hack" is still in there in exactly the same way as before, only now it isn't a hack, but the default value of the configurable line height.

litoj: > Just for example, tomorrow someone will create an issue that when setting the height, there is empty space at the top and cut off letters at the bottom

And that is precisely why the line height factor is adjustable like this. You would solve it simply by increasing the line height factor. The base height must be standardized, otherwise fonts would be in total chaos and also wouldn't be able to either print the bottom or top of their letters anywhere. So the 1/6 offset is the same for every font meaning this cannot ever happen in the first place.

 But just for the case where this were to happen, you would simply increase the line height factor. The text will always stay centered, so no problem there.

An explanation of why your hack worked is simple math - 1/3 is the total line height the font gives for letters that are out-of-line. Half of it is for the top, half for the bottom (-> 1/6). And by adding a 1/3 horizontal padding, you made it appear centered. I've just taken it further and generalized this math to add the spacing freely and account for the 1/6 from the bottom, to keep the text in center.

artemsen: Your reasoning is based on constants based on your assumptions.
If you do it using FontFree structures and constants, I'll come back to this issue.

artemsen: > It's not like the font cannot cause a crash ever - try using LANG=C and putting a  star emoji at the end of a line (I used it to mark the units of Rating). It will also crash, but due to the width. The problem there is with how weird locales clearly are, because using LANG=en_US.UTF-8 or any other lang works just fine.

Ooops.... I forgot to remove [debug code](https://github.com/artemsen/swayimg/commit/02cba77800b38ddf51b677a48244d849e65bbc1a). But it is not about font =)
Thank you!

PR #423, opened 2026-03-17
feat(text): line height configuration

litoj:
A rework of #422 to rely on data from FreeType instead of assuming font proportions.

### DIFF
--- a/src/font.cpp
+++ b/src/font.cpp
@@ -11,14 +11,14 @@
 #include <memory>
 #include <utility>
 
-constexpr const char *DEFAULT_FONT = "monospace"; // default font face
+constexpr const char* DEFAULT_FONT = "monospace"; // default font face
 
-constexpr size_t POINT_FACTOR = 64; // default points per pixel (26.6 format)
+constexpr size_t POINT_FACTOR = 64;  // default points per pixel (26.6 format)
 constexpr size_t MAX_TEXT_LEN = 120; // max length of text line (characters)
 
 /** Font config wrapper.*/
 class FontConfig {
-  public:
+public:
     using FcConfigPtr = std::unique_ptr<FcConfig, decltype(&FcConfigDestroy)>;
     using FcPatternPtr =
         std::unique_ptr<FcPattern, decltype(&FcPatternDestroy)>;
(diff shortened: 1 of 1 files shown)


src/font.cpp
```diff
@@ -157,10 +163,21 @@ Pixmap Font::render(const std::string& text)
...
+    const size_t height = height_base * this->height_factor;
+
+    // Centered line spacing scaling done by adding the extra space
+    // not used to print the char equally divided to the top and bottom
+    // (same for the horizontal padding to keep the corners on a 45° diagonal)
+
+    // Descender letters (gpqy) use some extra space at the bottom (saved as
+    // a negative value).
+    const size_t hpadding = height_base * (this->height_factor - 1) / 2 -
+        ft_face->descender / (short)POINT_FACTOR;
```
artemsen, line 175: This decrease in accuracy (cast double to short int) is a clear sign of adjustments to achieve the desired result.

artemsen (commented):
When I talked about FreeType structures and constants, I meant using the glyph description to calculate the position, not adjusting the calculations to the result.
Please stop this brute force.

src/font.cpp
litoj: I actually don't understand what you're talking about. I am casting a size_t into a short, because otherwise the `ft_face->descender` gets converted to a size_t, which cannot be negative, so the whole thing breaks.

My necessity to cast a natural into a whole number has absolutely nothing to do with some double that at maximum shifts the whole thing by 1px

litoj:
> When I talked about FreeType structures and constants, I meant using the glyph description to calculate the position, not adjusting the calculations to the result. Please stop this brute force.

What brute force are you talking about. I literally did what you wanted - I used the information about the font alignment that the font provides. There is no more correct way than the way I did it in (apart from maybe casting everything to a double to avoid the 1px potential error).

The idea you just described I really cannot understand. But I assume you thought each letter would be at a different height? Which is obviously not true. So I beg you, think about the code again, please.

artemsen:
Your PR:
image

Master branch:
image

You've simply tweaked the formula to suit your fonts.
It's not a universal solution.

litoj:
I am still working on the formula. Which font is this, so that I can test it?

artemsen:
> I am still working on the formula. Which font is this, so that I can test it?

Let's leave my font unknown so that you don't have to adjust the result to it.

artemsen:
@litoj FYI: https://github.com/artemsen/swayimg/commit/65fa4e6507e2d6f1a8b005f7630a8f1615d27de0

PR #472, opened 2026-04-15
gallery: customizable rows and columns

awsms:
Add Lua API to configure gallery columns and rows explicitly.

When fixed dimensions are set, calculate thumbnail cell width and height from the available window space instead of forcing square auto-filled tiles. This allows wide contact-sheet style layouts while keeping the previous automatic square grid behavior when rows and columns are set to 0.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -166,7 +166,9 @@ or in the file `/usr/share/swayimg/example.lua` after installing the program.
   * [swayimg.gallery.set_aspect](#swayimggalleryset_aspect): Set thumbnail aspect ratio
   * [swayimg.gallery.get_thumb_size](#swayimggalleryget_thumb_size): Get thumbnail size
   * [swayimg.gallery.set_thumb_size](#swayimggalleryset_thumb_size): Set thumbnail size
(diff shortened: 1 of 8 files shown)


artemsen:
The idea is certainly cool, but honestly, I don't see the need to crop the thumbnails.
If we keep the thumbnails square, the layout can be changed dynamically via lua, without requiring an API extension:
```lua
COLS = 2
ROWS = 3
PADDING = 10

swayimg.gallery.set_padding_size(PADDING)

swayimg.on_window_resize(function()
  if swayimg.get_mode() == 'gallery' then
    local wnd = swayimg.get_window_size()
    local tw = (wnd.width - COLS * PADDING) / COLS
    local th = (wnd.height - ROWS * PADDING) / ROWS
    swayimg.gallery.set_thumb_size(math.floor(math.min(tw, th)))
  end
end)
```

Also, PR does not take into account the aspect ratio and breaks the display of thumbnails in "keep" mode:
image
In addition to this, a new problem arises - now we have two different and conflicting approaches for setting the size of miniatures.

Frankly, I don't see any practical need to extend the Lua API yet.

PR #479, opened 2026-04-27
fix(#477): always print full error stacktrace

### LINKED ISSUE #477: Errors don't always contain full stacktrace
Errors that occur when a keymapping is pressed or during initialization and most other points print just the line where the error happened (useful only in trivial situations).

It would be extremely useful if the errors were printed in full detail like it is with errors occuring during config loading.

I've been struggling with this for a long time now, but I've always discovered a different bug at the same time or had a good guess for where the error could be coming from. Not this time however :/.

litoj:
Resolves #477

Note: this was AI aided so it certainly is possible that the types used in exec_lua could be improved, but I don't understand luabridge well enough to be able to tell if holding a simple pointer of the debug.traceback would work at all, so I leave it as is and up for debate

### DIFF
--- a/src/luaengine.cpp
+++ b/src/luaengine.cpp
@@ -270,9 +270,10 @@ void LuaEngine::initialize(const std::filesystem::path& config)
             const char* msg = lua_tostring(lua_state, -1);
             print_error("Failed to load config file: {}", msg ? msg : "<?>");
         } else {
-            const auto chunk = luabridge::LuaRef::fromStack(lua_state, -1);
+            const luabridge::LuaRef chunk =
+                luabridge::LuaRef::fromStack(lua_state, -1);
             lua_pop(lua_state, 1);
-            exec_lua(&chunk);
+            execute(&chunk);
         }
     }
 }
@@ -285,9 +286,10 @@ void LuaEngine::execute(const std::string& script)
         print_error("Failed to load script line: {}",
                     lua_tostring(lua_state, -1));
     } else {
-        const auto chunk = luabridge::LuaRef::fromStack(lua_state, -1);
+        const luabridge::LuaRef chunk =
+            luabridge::LuaRef::fromStack(lua_state, -1);
         lua_pop(lua_state, 1);
-        exec_lua(&chunk);
+        execute(&chunk);
     }
 }
 
@@ -302,11 +304,10 @@ void LuaEngine::bind_root_api()
         .addFunction("set_mode",
                      [this](const std::string& name) {
                          const auto mode = name_to_type(appmodes, name.c_str());
-                         if (mode.has_value()) {
-                             Application::self().set_mode(mode.value());
-                         } else {
-                             print_error("Invalid mode: {}", name);
+                         if (!mode.has_value()) {
+                             raise_error("Invalid mode: {}", name);
                          }
+                         Application::self().set_mode(mode.value());
                      })
         .addFunction("get_mode",
                      []() {
@@ -336,10 +337,9 @@ void LuaEngine::bind_root_api()
             "set_window_size",
             [this](const size_t width, const size_t height) {
                 if (!width || !height || width > 32767 || height > 32767) {
-                    print_error(
-                        "Invalid arguments ({}, {}) for {} .set_window_size ",
+                    raise_error(
+                        "Invalid arguments ({}, {}) for {}.set_window_size",
                         width, height, NS_SWAYIMG);
-                    return;
                 }
                 Ui* ui = Application::get_ui();
                 if (ui) {
@@ -356,15 +356,14 @@ void LuaEngine::bind_root_api()
             "on_window_resize",
             [this](const luabridge::LuaRef& cb) {
                 if (!cb.isFunction()) {
-                    print_error("Invalid argument for {}.on_window_resize: "
+                    raise_error("Invalid argument for {}.on_window_resize: "
                                 "expected function, but got {}",
-                                NS_SWAYIMG, cb.tostring().c_str());
-                } else {
-                    const luabridge::LuaRef* ref = add_ref(&cb);
-                    Application::self().subscribe_window_resize([this, ref]() {
-                        exec_lua(ref);
-                    });
+                                NS_SWAYIMG, cb.tostring());
                 }
+                const luabridge::LuaRef* ref = add_ref(&cb);
+                Application::self().subscribe_window_resize([this, ref]() {
+                    execute(ref);
+                });
             })
         .addFunction("get_mouse_pos",
                      []() {
@@ -422,15 +421,14 @@ void LuaEngine::bind_root_api()
         .addFunction("on_initialized",
                      [this](const luabridge::LuaRef& cb) {
                          if (!cb.isFunction()) {
-                             print_error(
+                             raise_error(
                                  "Invalid argument for {}.on_initialized: "
                                  "expected function, but got {}",
-                                 NS_SWAYIMG, cb.tostring().c_str());
-                             return;
+                                 NS_SWAYIMG, cb.tostring());
                          }
                          const luabridge::LuaRef* ref = add_ref(&cb);
                          Application::self().on_init_complete = [this, ref]() {
-                             exec_lua(ref);
+                             execute(ref);
                          };
                      })
         .addFunction("enable_antialiasing",
@@ -448,13 +446,13 @@ void LuaEngine::bind_root_api()
                    const std::unordered_map<std::string, bool>& params) {
                 ImageFormat* fmt = FormatFactory::self().get(name.c_str());
                 if (!fmt) {
-                    print_error("Unsupported format {} for {}.on_initialized",
-                                name, NS_SWAYIMG);
-                    return;
+                    raise_error(
+                        "Unsupported format {} for {}.set_format_params", name,
+                        NS_SWAYIMG);
                 }
                 if (!fmt->set_params(params)) {
-                    print_error("Unsupported parameters for format {} in "
-                                "{}.on_initialized",
+                    raise_error("Unsupported parameters for format {} in "
+                                "{}.set_format_params",
                                 name, NS_SWAYIMG);
                 }
             })
@@ -470,10 +468,9 @@ void LuaEngine::bind_root_api()
             "set_dnd_button",
             [this](const std::string& button) {
                 std::optional<InputMouse> input = InputMouse::load(button);
-                if (!input) {
-                    print_error("Invalid button for {}.set_drag_button: {}",
+                if (!input.has_value()) {
+                    raise_error("Invalid button for {}.set_drag_button: {}",
                                 NS_SWAYIMG, button);
-                    return;
                 }
                 Application::self().sparams.dnd = input.value();
             })
@@ -511,11 +508,10 @@ void LuaEngine::bind_imagelist_api()
                      [this](const std::string& name) {
                          const auto order =
                              name_to_type(ilorders, name.c_str());
-                         if (order.has_value()) {
-                             ImageList::self().set_order(order.value());
-                         } else {
-                             print_error("Invalid image list order: {}", name);
+                         if (!order.has_value()) {
+                             raise_error("Invalid image list order: {}", name);
                          }
+                         ImageList::self().set_order(order.value());
                      })
         .addFunction("enable_reverse",
                      [](const bool enable) {
@@ -598,21 +594,6 @@ void LuaEngine::bind_text_api()
         .endNamespace();
 }
 
-// generate lambda for checking if required mode is active
-auto mk_mode_checker(lua_State* L, const char* mode_name)
-{
-    auto mode = name_to_type(appmodes, mode_name).value();
-    return [L, mode, mode_name](const char* fname) {
-        if (Application::self().get_mode() == mode) {
-            return true;
-        }
-
-        throw luabridge::raise_lua_error(
-            L, "Unable to execute %s.%s.%s - mode not active", NS_SWAYIMG,
-            mode_name, fname);
-    };
-}
-
 void LuaEngine::bind_viewer_api(const char* name)
 {
     assert(strcmp(name, NS_VIEWER) == 0 || strcmp(name, NS_SLIDESHOW) == 0);
@@ -623,27 +604,32 @@ void LuaEngine::bind_viewer_api(const char* name)
     bind_appmode_api(name);
 
     // check if required mode is active
-    auto assert_active = mk_mode_checker(lua_state, name);
+    auto ensure_active = [this, mode, name](const char* fname) {
+        if (Application::self().current_mode() == mode) {
+            return;
+        }
+        raise_error("Unable to execute {}.{}.{} - mode not active", NS_SWAYIMG,
+                    name, fname);
+    };
 
     luabridge::getGlobalNamespace(lua_state)
         .beginNamespace(NS_SWAYIMG)
         .beginNamespace(name)
         .addFunction(
             "switch_image",
-            [this, assert_active, mode, name](const std::string& dname) {
-                assert_active("switch_image");
+            [this, ensure_active, mode, name](const std::string& dname) {
+                ensure_active("switch_image");
                 const auto dir = name_to_type(ildirs, dname.c_str());
-                if (dir.has_value()) {
-                    mode->open(dir.value());
-                } else {
-                    print_error(
+                if (!dir.has_value()) {
+                    raise_error(
                         "Invalid argument \"{}\" for {}.{}.switch_image", dname,
                         NS_SWAYIMG, name);
                 }
+                mode->open(dir.value());
             })
         .addFunction("open",
-                     [assert_active, mode](const std::string& path) {
-                         assert_active("open");
+                     [ensure_active, mode](const std::string& path) {
+                         ensure_active("open");
                          ImageEntryPtr entry =
                              Application::self().add_image_entry(path);
                          if (!entry) {
@@ -655,8 +641,8 @@ void LuaEngine::bind_viewer_api(const char* name)
                          }
                      })
         .addFunction("get_image",
-                     [this, assert_active, mode]() {
-                         assert_active("get_image");
+                     [this, ensure_active, mode]() {
+                         ensure_active("get_image");
                          const ImagePtr image = mode->current_image();
                          luabridge::LuaRef tbl = entry_to_table(*image->entry);
                          tbl["format"] = image->format;
@@ -672,13 +658,13 @@ void LuaEngine::bind_viewer_api(const char* name)
                          return tbl;
                      })
         .addFunction("reload",
-                     [assert_active, mode]() {
-                         assert_active("reload");
+                     [ensure_active, mode]() {
+                         ensure_active("reload");
                          mode->reload();
                      })
         .addFunction("reset",
-                     [assert_active, mode]() {
-                         assert_active("reset");
+                     [ensure_active, mode]() {
+                         ensure_active("reset");
                          mode->reset();
                      })
         .addFunction("get_scale",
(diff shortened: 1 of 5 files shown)


artemsen:
Please do not submit pull requests unless you understand what is implemented in it.
It didn't even pass the unit tests.

```
[----------] 1 test from LuaEngineTest
[ RUN      ] LuaEngineTest.Load
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
1/1 swayimg:Unit tests FAIL            0.11s   killed by signal 6 SIGABRT


Summary of Failures:

1/1 swayimg:Unit tests FAIL            0.11s   killed by signal 6 SIGABRT

Ok:                0   
Fail:              1   
```

litoj:
So, just to clear some things up.
- I am trying to fix a problem. Fixing a problem often doesn't work on the first try - that's not a reason to always immediately close the PR when not everything works as expected.
- I am not used to meson, etc. So running tests is just something I have to do manually, which I always forget (because I'm used to leaving testing to the pipeline which has the results that actually count).
- It feels wrong to be expected to have everything correct on the first try (mistakes are in human nature) especially when even the tests are wrong and undeterministic (relying on LC_COLLATE settings to make the test pass correctly, yet even then the pipeline tests get different results than on my machine)
- The test results are wrong in their own anyway. What failed? Well, after studying each line of the log output I finally noticed a SIGSEGV, but if each test ran isolated then the test results would be able to work with that instead of crashing everything, leaving me with unclear output that makes it easy to overlook the actual error. It would be better if tests were run in debug mode so that we'd at least get a stacktrace to the error. Also, the test results say Fail: 1, Ok: 0, as if there was only 1 test - so that's wrong too.

With these issues already in the project hindering its further development, why be so quick to reject any attempts that are putting in the effort to overcome those issues in order to actually fix a problem in the project?

I submit only the code I understand or believe I do. That doesn't mean, however, that I understand every detail of the underlying libraries that are being used. I am no magician, which is why I can only do _my_ best to try to contribute. It's a process anyway - not an instant unchangeable state.

Which is why I'd like to ask you to reopen this PR so I can work on it and try to figure out the magical issue that happens only in tests and not when actually using the app (as is often the case as luck would have it :D), so that another issue can eventually get fixed instead of being stuck in a state where a solution is being worked on but instantly rejected unless it seems perfect - which from the perspective of a different person is practically guaranteed to always mean sth different, resulting in only wasted time on both sides and no change benefiting the end user.

artemsen:
> I am trying to fix a problem. Fixing a problem often doesn't work on the first try - that's not a reason to always immediately close the PR when not everything works as expected.

I agree. But PR needs to be checked before submission. At the very least, it should pass CI.

> I am not used to meson, etc. So running tests is just something I have to do manually, which I always forget (because I'm used to leaving testing to the pipeline which has the results that actually count).

We have CI for this. Enable actions on your fork.

> It would be better if tests were run in debug mode so that we'd at least get a stacktrace to the error. Also, the test results say Fail: 1, Ok: 0, as if there was only 1 test - so that's wrong too.

Test is for testing. If you need to debug - build it locally in debug mode it run under GDB/LLDB/etc. (but even in debug mode it will not print stack trace to console).

> With these issues already in the project hindering its further development, why be so quick to reject any attempts that are putting in the effort to overcome those issues in order to actually fix a problem in the project?

Because you're sending AI slop and don't even check it. I don't want to waste my time analyzing auto-generated code.

> Which is why I'd like to ask you to reopen this PR

Okay, but please send the results when they are ready for review, and not just "see what AI can do" =)

litoj:
> We have CI for this. Enable actions on your fork.

I have never worked with github CI. I had no idea you have actions on commits as well - not just for PRs. I will definitely try that next time, thank you!

litoj:
> Because you're sending AI slop and don't even check it. I don't want to waste my time analyzing auto-generated code.

This is exactly what I was afraid your reasons were. Had I not mentioned the word `AI` things would have been different, despite the code logic and ideas being mine, not from the AI (I used it primarily to fill in the correct types and syntax for lambda definitions). But understandably it is difficult to tell when the mention of AI is as offputting as it has become or also when the programmer (me) is as bad in the particular language as an AI would be.

I need to ephasize this before saying what I'm about to say: _this is just a personal revelation and not an attempt at lecturing_ - Admitting any use of AI in a PR to any project results in the PR being automatically considered the lowest quality no matter what the AI was actually used for. This incentivises people to hide their sources instead of being forwardcomming and acknowledging that not a full 100% of the PR was their work. Even though it is true in the jobmarket as well, that people that admit their faults get nowhere, it is sad to see that because of AI the same can nowadays be said about FOSS in general, instead of being considered a polite suggestion that the author welcomes suggestions to improve the code.

litoj:
> Because you're sending AI slop and don't even check it. I don't want to waste my time analyzing auto-generated code.
... send the results when they are ready for review, and not just "see what AI can do" =)

Just to reiterate - From what I understand in this case it wasn't me, who didn't even check the code. It is also very sad to see one's work being considered all AI slop just because of the pure mention of AI, no matter how untrue it may be.

I committed good working changes from the start. But when fixing the formatting, I also added another minor change that was meant to just get rid of the duplication of logic for handling mouse/key/signal events, which just added another lambda wrap to everything. I suppose I forgot to recompile because it worked for after that as well. -- That's all - just a normal human mistake like they happen every day. Nothing to do with AI.

Anyways, I reverted those changes and just reformatted the file and I'll make a separate PR for that simplification at some other point. This PR is as ready as can be, but suggestions for changes are still welcome.

src/luaengine.cpp
```diff
@@ -19,6 +19,8 @@
 #include <cstring>
 #include <optional>
 
+#include "lua.h"
```
artemsen: This line is not necessary.

src/luaengine.cpp
```diff
@@ -213,6 +215,7 @@ LuaEngine::~LuaEngine()
         for (auto& it : refs) {
             delete it;
         }
+        traceback_fn = nullptr;
```
artemsen: This line is not necessary.

src/luaengine.hpp
```diff
@@ -20,6 +20,8 @@
 #include <format>
 #include <vector>
 
+#include "lua.h"
```
artemsen: This line is not necessary.

src/luaengine.hpp
artemsen: I suppose this stacktrace printer is not needed anymore.

artemsen (commented):
Also, please fix the [commit message](https://gist.github.com/turbo/efb8d57c145e00dc38907f9526b60f17):
```
lua: Print full stacktrace

Resolves ...

Signed-off-by: ...
```

meson.build
```diff
@@ -404,5 +404,5 @@ if get_option('tests').enabled()
         dependencies: dependencies + gtest,
         cpp_args: '-DTEST_DATA_DIR="' + meson.current_source_dir() + '/test/data"',
     )
-    test('Unit tests', test_exe)
+    test('Unit tests', test_exe, env: ['LC_COLLATE=en_US.UTF-8'])
```
artemsen: This line is not necessary.

litoj: this line _is_ necessary to make the sorting work with any other system locale.

Why do you think I'd touch a completely unrelated part of the codebase otherwise, if it wasn't annoying me in every test run.

artemsen: > this line is necessary to make the sorting work with any other system locale.

Locales are configured in docker container or on local host.

> Why do you think I'd touch a completely unrelated part of the codebase otherwise, if it wasn't annoying me in every test run.

Please don't add solutions of your local problems to the source code.

litoj: It is not just my problem. It is a problem for everyone not using this particular locale. Tests are supposed to be consistent. That means independent of local host configuration. Therefore the tests should be configured in a way that doesn't require every user to adjust their system to the tests, but rather the tests to eliminate the influence of user system configuration.

Literally the whole point of testing - get reproducible results about the _code_, _not_ to detect which locale is the user running on their system.

Why are you always so rejective of the simplest of improvements?

litoj: Think about it the other way - if it is a problem for literally everyone not using US locale is it really that everyone has local problems, or is it that the tests are bad because they depend on external configuration that they can't influence and therefore are unreliable?

artemsen: > it is a problem for literally everyone not using US locale

I don't know where you got this from, but Swayimg doesn't require the US locale to work.
It works with _any_ utf-8 locale in Linux. Please, configure [your system](https://github.com/artemsen/swayimg/issues/407#issuecomment-4062673669) properly.

litoj: Swayimg indeed doesn't require the US locale. However, the alpha sorting test does, because the sorting differs depending on the `LC_COLLATE` env variable. This means that the alphabetical sorting tests may fail if the LC_COLLATE setting is not set to US.

And again, tests should work correctly for all environments where the application may run in. So if they fail in an environment with a locale that affects the sorting order, then that means the tests are broken, because they suggest the application won't work in an environment where the application clearly does work.

I do agree with you that removing the locale setting from the tests entirely is not the best way to do it, but we are not setting the locale to US, just the sorting order, which is assumed to be US in all tests in the first place (hence why this fixes it).

If you were testing locale, you'd have to have tests that run the testing suite with different locales, but this isn't being done here, which means that locale is just a loose variable that breaks the tests instead of being taken as a core part of the tests.

The problem is that is is impossible to fix this testing error via any other means because the locale is used as a function-static variable: https://github.com/artemsen/swayimg/blob/master/src/imagelist.cpp#L26

This makes it impossible to mock or change other than before any comparison is done, therefore before the tests are run.

PS: Sorry for the delay I seem to have missed the email notification.

litoj:
Hi, I've just spent the whole day trying to find ways to avoid duplicating the `check_active` logic. Unfortunately there is no way to allow making a wrapper that would check the mode first, and if not having to throw an error it would call the actual function. But with luabridge clever argument parsing it is not possible to wrap the lambda without a heap of type hacking around (if possible at all). So I decided to put the lambda for generating the api errors outside the mode bindings instead to avoid at least the duplication of the handler code.

meson.build
artemsen: > Swayimg indeed doesn't require the US locale.

I still don't understand why you decided that the US locale is mandatory. Because it's specified in CI? You can set any other utf8 locale there and it will work (on Linux, but not on FreeBSD). This test uses the [Russian alphabet](https://github.com/artemsen/swayimg/blob/164e22ee920411617d890078b51402a08220e975/test/imagelist_test.cpp#L177), so by your logic, we should use ru_RU. However, this is not required.

> The problem is that is is impossible to fix this testing error via any other means because the locale is used as a function-static variable

I doubt that's the problem. We don't need to support dynamic locale changes at runtime.
It's deliberately made static, as this is a frequently called function, and each locale reinitialization results in a significant performance penalty.

artemsen:
I decided to put the lambda for generating the api errors outside the mode bindings instead to avoid at least the duplication of the handler code

Please don't mix different issues in a single pull request. We'll never close it that way =)

src/luaengine.cpp
```diff
@@ -594,6 +598,25 @@ void LuaEngine::bind_text_api()
         .endNamespace();
 }
 
+// generate lambda for checking if required mode is active
+auto mk_mode_checker(lua_State* L, const char* mode_name)
```
artemsen: Sorry, but I don't like this approach. A little copy-paste is better than such convoluted logic with lambda function generation.

litoj: I am sorry, but are you seriously calling _that_ a compicated logic? :D

It is literally the same lambda as it was before, the only difference being that I am defining an extra parameter (the mode). Barely half a line has changed.

If the capture definition is what looks bad, I do agree, but that one we can easily transform into:
```cpp
    auto mode = name_to_type(appmodes, mode_name).value();
    return [L, mode_name, mode](const char* fname) {
```

Would that change suffice?

litoj: I've found an even simpler way to do it. Now I think it is really as simple as it can be.

litoj:
> Please don't mix different issues in a single pull request. We'll never close it that way =)

I know, you're right. I just couldn't resist because I was already having to change most of the same parts to use the throw_error(). So it felt like a natural fit to also refactor the duplication of the check_active into the assert_active maker. I originally tried to edit both of them, but it was so annoying, because I always thought I had it right so I copied it over to the other lambda, and then I thought of yet another improvement to its implementation. It was just so much simpler and more practical to have the actual handler logic in one place and let it be called from wherever necessary.

artemsen:
> I am sorry, but are you seriously calling _that_ a compicated logic?

Calling a lamda function that is generated by another static function, and which is also capable of interrupting the calling function, is complex logic. Add to this the need to keep two contexts in mind when analyzing (the parameters passed to the lambda function and the parameters used when creating it) and you get clear evidence of poor design.
So yes, I think it's complex logic. This code is easy to write but hard to read and maintain, so you'll have to come up with something else.

> I originally tried to edit both of them, but it was so annoying, because I always thought I had it right so ...

It's a never-ending process. If it weren't for the locale changes, I would have merged your code a week ago and you would have moved on.

And once again about localization.
As I said [earlier](https://github.com/artemsen/swayimg/pull/479#discussion_r3176843958), Swayimg doesn't require the US locale. The test uses the [Russian alphabet](https://github.com/artemsen/swayimg/blob/164e22ee920411617d890078b51402a08220e975/test/imagelist_test.cpp#L177), so by your logic, we should use ru_RU.
But I'm willing to discuss this issue again in a separate PR, provided you find a way to reproduce the error on more than just your local machine (for example, in a Docker container).
Please remove the code from this PR that is not relevant to its title.

litoj:
> But I'm willing to discuss this issue again in a separate PR, provided you find a way to reproduce the error on more than just your local machine (for example, in a Docker container).

I can reproduce it even on your machine. Just run the tests with LC_COLLATE or LC_ALL set to C or cs_CZ.UTF-8. One will fail the Unicode test, the other one will fail the Alpa tests.

I'll try to undo the changes you wish to discuss separately and make a new pr.

artemsen:
> cs_CZ.UTF-8

My apologies, you were right - this is the specificity of the Czech locale. It treats numbers and characters based
on their Unicode codepoints, not numeric value, and sorts numbers after letters.
Fixed in b659df7.

src/luaengine.cpp
```diff
@@ -261,23 +269,25 @@ void LuaEngine::initialize(const std::filesystem::path& config)
...
+        } else {
+            const auto chunk = luabridge::LuaRef::fromStack(lua_state, -1);
+            lua_pop(lua_state, 1);
+            execute_lua(&chunk);
         }
     }
 }
 
-void LuaEngine::execute(const std::string& script)
+void LuaEngine::execute_string(const std::string& script)
```
artemsen: You shouldn't have renamed it. What does it mean to execute a string? It executes code, not a string.
In c++ you can have multiple functions with the same name but different arguments. So, thay are all csn be just "execute", for string, ref or file.

src/luaengine.cpp
```diff
@@ -261,23 +269,25 @@ void LuaEngine::initialize(const std::filesystem::path& config)
...
     assert(lua_state);
 
     if (luaL_loadstring(lua_state, script.c_str()) != LUA_OK) {
         print_error("Failed to load script line: {}",
                     lua_tostring(lua_state, -1));
-    } else if (lua_pcall(lua_state, 0, 0, 0) != LUA_OK) {
-        print_error("Failed to execute script line: {}",
-                    lua_tostring(lua_state, -1));
+    } else {
+        const auto chunk = luabridge::LuaRef::fromStack(lua_state, -1);
```
artemsen: Please do not use `auto` if a specific type can be specified. It make code more readable.

src/luaengine.cpp
```diff
@@ -380,7 +384,7 @@ void LuaEngine::bind_root_api()
                          } else {
                              std::optional<bool>& fullscreen =
                                  Application::self().sparams.fullscreen;
-                             if (!fullscreen.has_value()) {
```
artemsen, line 383: By doing this you have made the code less readable.
Now the value check looks like "if not fullscreen", but in fact it is "if fullscreen is not specified".
Please revert it back.

src/luaengine.cpp
```diff
@@ -601,41 +599,36 @@ void LuaEngine::bind_viewer_api(const char* name)
     Viewer* mode =
         strcmp(name, NS_VIEWER) == 0 ? &Viewer::self() : &Slideshow::self();
 
-    bind_appmode_api(name);
+    bind_appmode_api(name, mode);
 
     // check if required mode is active
-    auto check_active = [this, mode, name](const char* fname) {
+    auto assert_active = [this, mode, name](const char* fname) {
```
artemsen: Not the best rename.
Assert in C++ assumes checking in debug mode and is disabled in release.

src/luaengine.hpp
```diff
@@ -87,21 +94,25 @@ class LuaEngine {
...
+        Text::self().set_status(message.substr(0, message.find_first_of('\n')));
+    }
+
+    /**
+     * Throw an exception with Lua error info.
+     * @param fmt format description
+     * @param ... format arguments
+     */
+    template <typename... Args>
+    void throw_error(const std::format_string<Args...> fmt,
```
artemsen: Bad name.This function in fact doesn't throw exception, but uses long jump.
Lets rename it to `raise_error`.
It was also helpful to mark it as `[[noreturn]]`.

src/luaengine.cpp
```diff
@@ -1073,18 +1026,8 @@ void LuaEngine::bind_gallery_api()
         .endNamespace();
 }
 
-void LuaEngine::bind_appmode_api(const char* name)
+void LuaEngine::bind_appmode_api(const char* name, AppMode* appmode)
```
artemsen: Now these two parameters mean the same thing, but they conflict with each other.
With this change, you added the ability to create bindings for a mode with an incorrect namespace. Please revert it back.

src/luaengine.cpp
litoj: I thought it would be better to distinguish execute_lua from execute, but I can make both just be `execute`, allright

src/luaengine.cpp
litoj: why duplicate code logic instead of passing in the correct arguments?

With your approach adding in a new mode requires changing this method despite the method having no reason to need to know which particular appmode it is because it works the same way for all of them. You are using cpp but writing C code.

The correct way is to make appmode advertise its name instead of using mode name arrays and enums and conversions etc.

src/luaengine.cpp
litoj: literally every saved result of `name_to_type` or the opposite you use a `const auto ...`. The types are mostly obvious in all cases anyway.

For primitive single use cases it really doesn't matter, but sure.

src/luaengine.cpp
litoj: I can make a separate PR for this CQ improvement, if you want

artemsen: > adding in a new mode

I can't even find what other mode there might be.

> correct way is to make appmode advertise

This is how it happened historically. But there are only 3 modes, new ones are unlikely to appear, so this is not a problem.
Let's leave it as is for now; reworking this mechanism "correctly" will affect the reworking of global interfaces, but it's not worth doing that yet.

src/luaengine.cpp
```diff
@@ -953,13 +914,12 @@ void LuaEngine::bind_slideshow_api()
 void LuaEngine::bind_gallery_api()
 {
     // check if required mode is active
-    auto check_active = [this](const char* name) {
-        if (Gallery::self().is_active()) {
-            return true;
+    auto ensure_active = [this](const char* fname) {
+        if (!Gallery::self().is_active()) {
```
litoj, line 918: @artemsen Sorry, I seem to have confused myself when trying to revert my mode detection. There should be no negation in here. Currently I made it impossible to switch images in gallery

src/luaengine.cpp
litoj: > I can't even find what other mode there might be.

Well, I've already made two with my lua api and have ideas for a third one :D. One could be obviously a help mode, but I also hacked the gallery mode into working like a two-pane view. Having two images next to each other to be able to look at them and decide which one is better and which one can be deleted. Ofc, this is a feature of image editors, but there ceartainly is place for more modes :)

src/luaengine.cpp
artemsen: Fixed.
Thank you!

src/luaengine.cpp
artemsen: > two-pane view

Not a bad idea, by the way. I'll think about it.

PR #486, opened 2026-05-19
doc: small fix

yushisora:


artemsen:
This is an auto-generated file from MD, and it also needs to be changed:
https://github.com/artemsen/swayimg/blob/225d37833c1bfdeb67e8729f72b360f32f87c5f1/USAGE.md?plain=1#L84

yushisora:
done

Issue #481, opened 2026-05-03
Search using regular expressions and EXIF ​​metadata


tuf612:
Would it be possible to add a search feature to swayimg’s gallery mode? The search box would not be permanently visible.

instead, it would be invoked on demand via configurable key bindings — similar to how searching works in Vim or less.

Typical key bindings for this could be:

- `/` to open the search box and start a forward search through the file list, **starting from the currently selected item**
- `?` to open the search box and start a backward search through the file list, **starting from the currently selected item**

Once the search box appears, users could type a pattern and see real-time highlighting of matching images as they type.

In terms of search scope, it would be extremely powerful if the matching went beyond filenames and supported regex matching against EXIF metadata (e.g., camera model, lens, date, ISO, focal length, etc.).

This would turn the gallery into a very efficient tool for navigating large, metadata-rich image collections.

Thank you for considering this!

artemsen:
I doubt I'll ever take it on, too much UI. But you never know.

litoj:
Hi, I like your idea, so I decided to implement it. It took a couple of days, but only thing that's missing are the mappings. I've set `<C-j>/<C-k>` as the default instead, but it's easy to change. Anyways, enough talking. Does this hit the mark?

https://github.com/user-attachments/assets/5b1e5b56-7f84-4525-b490-6ff0ff6a30be

There are a couple of problems currently: displaying and hiding images requires changing the list contents, not the content visibility. So this makes reordering slow.

The main problem is that we don't have a way to do bulk changes. @artemsen could we get a `swayimg.imagelist.set()` method for bulk changes? Currently most of the time is wasted on exchanging locks and when loading thumbnails in gallery while trying to filter, removing 1800 out of 2000 images takes 10s, but only 3 when thumbnails are done. Locking it for longer to make all the changes at once would be a serious performance boost.

Second drawback is the exif data - currently, exif data is available only in viewer mode and only for the current image, so I had to add a cpp module that loads the exif data for any path you give it. Surprisingly, even for 2000 files on a single-thread load in under a second, so this is more about ease of use than any practical issues.

Third to solve was to determine a way of filtering - you'd probably want to choose the exif data to filter by making the filter be an expression. I allowed for standard operators and then the `:` code operator for maximum flexibility.

It's part of my [swi.swi](https://github.com/litoj/swi.swi) plugin if you'd like to try it. All suggestions for improvements are welcome.

artemsen:
[...]

PR #488, opened 2026-05-20
gallery: cache performance

litoj:
1st change in a series of changes making large imagelist changes more performant

relates to #481

I snuck in also the `clear_cache` api method that now allows the user to actually reload the thumbnails. This is useful in cases like large thumbnail size increase where the quality of the already generated thumbnails in memory would otherwise compromise the UX.

### DIFF
--- a/src/gallery.cpp
+++ b/src/gallery.cpp
@@ -346,20 +346,19 @@ void Gallery::window_redraw(Pixmap& wnd)
 
     const ImageEntryPtr current = layout.get_selected();
 
+    const auto& scheme = layout.get_scheme();
+    size_t selected_scheme_idx = 0;
+    size_t i = 0;
     // draw all exclude the currently selected
-    for (const auto& it : layout.get_scheme()) {
-        if (it.img != current) {
-            draw(it, wnd);
-        }
-    }
-
-    // draw only currently selected
-    for (const auto& it : layout.get_scheme()) {
+    for (const auto& it : scheme) {
         if (it.img == current) {
+            selected_scheme_idx = i;
+        } else {
             draw(it, wnd);
-            break;
+            i++;
         }
     }
+    draw(scheme[selected_scheme_idx], wnd);
 
     load_thumbnails();
     clear_invisible_thumbnails();
--- a/src/gallery.cpp
+++ b/src/gallery.cpp
@@ -362,7 +362,7 @@ void Gallery::window_redraw(Pixmap& wnd)
     }
 
     load_thumbnails();
-    clear_thumbnails();
+    clear_invisible_thumbnails();
 }
 
 void Gallery::handle_mmove(const InputMouse&, const Point& pos, const Point&)
@@ -383,13 +383,7 @@ void Gallery::handle_imagelist(const ImageListEvent event,
     if (event == ImageListEvent::Modify || event == ImageListEvent::Remove) {
         // remove entry from cache
         const std::scoped_lock lock(mutex);
-        auto it = std::find_if(cache.begin(), cache.end(),
-                               [&entry](const ThumbEntry& thumb) {
-                                   return entry == thumb.entry;
-                               });
-        if (it != cache.end()) {
-            cache.erase(it);
-        }
+        cache.erase(entry);
     }
 
     if (event == ImageListEvent::Remove && entry == layout.get_selected()) {
@@ -552,7 +546,7 @@ void Gallery::load_thumbnails()
     }
 }
 
-void Gallery::clear_thumbnails()
+void Gallery::clear_invisible_thumbnails()
 {
     const std::scoped_lock lock(mutex);
 
@@ -565,20 +559,19 @@ void Gallery::clear_thumbnails()
             visible_first > cache_size / 2 ? visible_first - cache_size / 2 : 1;
         const size_t store_max = visible_last + cache_size - cache_size / 2;
 
-        std::erase_if(cache, [store_min, store_max](const ThumbEntry& thumb) {
-            return thumb.entry->index < store_min ||
-                thumb.entry->index > store_max;
-        });
+        std::erase_if(cache,
+                      [store_min, store_max](
+                          const std::pair<ImageEntryPtr, Pixmap>& key_value) {
+                          return key_value.first->index < store_min ||
+                              key_value.first->index > store_max;
+                      });
     }
 }
 
 const Pixmap* Gallery::get_thumbnail(const ImageEntryPtr& entry)
 {
-    const auto it = std::find_if(cache.begin(), cache.end(),
-                                 [&entry](const ThumbEntry& th) {
-                                     return entry == th.entry;
-                                 });
-    return it == cache.end() ? nullptr : &it->pm;
+    const auto it = cache.find(entry);
+    return it == cache.end() ? nullptr : &it->second;
 }
 
 void Gallery::load_thumbnail(const ImageEntryPtr& entry)
@@ -591,28 +584,26 @@ void Gallery::load_thumbnail(const ImageEntryPtr& entry)
 
     const size_t thumb_size = layout.get_thumb_size();
 
-    ThumbEntry thumb;
-    thumb.entry = entry;
+    Pixmap pm;
 
     if (pstore_enable) {
-        thumb.pm = pstore_load(entry);
+        pm = pstore_load(entry);
     }
 
-    if (!thumb.pm) {
-        thumb.pm = FormatFactory::self().preview(entry, thumb_size,
-                                                 aspect == Aspect::Fill);
-        if (!thumb.pm) {
+    if (!pm) {
+        pm = FormatFactory::self().preview(entry, thumb_size,
+                                           aspect == Aspect::Fill);
+        if (!pm) {
             Application::self().add_event(AppEvent::FileRemove { entry->path });
         } else if (pstore_enable) {
-            pstore_save(entry, thumb.pm);
+            pstore_save(entry, pm);
         }
     }
 
     const std::scoped_lock lock(mutex);
 
-    if (thumb.pm) {
-        thumb.entry = entry;
-        cache.emplace_back(thumb);
+    if (pm) {
+        cache.insert_or_assign(entry, pm);
     }
     active.erase(entry);
 
--- a/src/gallery.hpp
+++ b/src/gallery.hpp
@@ -9,9 +9,9 @@
 #include "layout.hpp"
 #include "threadpool.hpp"
 
-#include <deque>
 #include <mutex>
 #include <set>
+#include <unordered_map>
 
 class Gallery : public AppMode {
 public:
@@ -168,9 +168,9 @@ private:
     void load_thumbnails();
 
     /**
-     * Remove unused thumbnails from the cache.
+     * Remove invisible thumbnails from the cache.
      */
-    void clear_thumbnails();
+    void clear_invisible_thumbnails();
 
     /**
      * Get thumbnail pixmap for specified image.
@@ -217,12 +217,11 @@ private:
     bool pstore_enable; ///< Use persistent storage for thumbnails
     std::filesystem::path pstore_path; ///< Persistent storage path
 
-    /** Thumbnail entry. */
-    struct ThumbEntry {
-        ImageEntryPtr entry;
-        Pixmap pm;
-    };
-    std::deque<ThumbEntry> cache;   ///< Thumbnails cache
+    /**
+     * Thumbnails cache.
+     * ImageEntry caching allows determining the current index of the thumbnail.
+     */
+    std::unordered_map<ImageEntryPtr, Pixmap> cache;
     std::set<ImageEntryPtr> queue;  ///< Loading queue
     std::set<ImageEntryPtr> active; ///< Currently loading
 
--- a/src/compositor.cpp
+++ b/src/compositor.cpp
@@ -355,7 +355,7 @@ public:
     static Rectangle get_focus()
     {
         Rectangle wnd;
-        int monitor_id;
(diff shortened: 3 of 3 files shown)


extra/swayimg.lua
```diff
@@ -723,6 +723,10 @@ function swayimg.gallery.enable_hover(enable) end
 ---@param size integer Cache size
 function swayimg.gallery.limit_cache(size) end
 
+---Clear thumbnails stored in memory cache.
+---Since 5.3.
+function swayimg.gallery.clear_cache() end
```
artemsen: Definitely no. You can't take the cache out, this is the program's internal workings.

litoj: clearing the cache is not allowed, but changing its size is? What is the distinction for why no to one but yes to another? What's the logic here?

It has no downside, simple implementation that works perfectly, and allows the user to change thumbnail size in runtime without having to do trickery with setting cache size to 0, scrolling to a new set of images to get the old cache cleared, scrolling back and then setting back the cache size, which is extremely annoying (especially when there aren't enough images on-screen to be able to unload them all) and has no technical reason to have to stay that way.

artemsen: > clearing the cache is not allowed, but changing its size is? What is the distinction for why no to one but yes to another?

Because setting cache parameters and manually modifying it are two different things.

> It has no downside

It has. Now, going forward, we need to take into account that the cache can be modified externally, which complicates the entire system.

> allows the user to change thumbnail size in runtime without having to do trickery with setting cache size to 0, scrolling to a new set of images to get the old cache cleared, scrolling back and then setting back the cache size, which is extremely annoying (especially when there aren't enough images on-screen to be able to unload them all) and has no technical reason to have to stay that way.

It's unclear to me why cache needs to be cleared manually at all.

litoj: What I haven't realized though, is that `swayimg.gallery.reload()` already does just that, so no need to add this after all.

Fixed!

src/gallery.hpp
```diff
@@ -170,7 +170,7 @@ class Gallery : public AppMode {
     /**
      * Remove unused thumbnails from the cache.
      */
-    void clear_thumbnails();
+    void clear_offsceen_thumbnails();
```
artemsen: There is no need for this renaming.

src/gallery.cpp
```diff
@@ -591,28 +585,25 @@ void Gallery::load_thumbnail(const ImageEntryPtr& entry)
...
         }
     }
 
     const std::scoped_lock lock(mutex);
 
-    if (thumb.pm) {
-        thumb.entry = entry;
-        cache.emplace_back(thumb);
+    if (pm) {
+        cache[entry] = pm;
```
artemsen: `cache.insert_or_assign(pm)` would be better choice, it avoids default construction overhead.

artemsen (changes requested):

artemsen:
Please follow [this](https://gist.github.com/turbo/efb8d57c145e00dc38907f9526b60f17) rules to create commit, or use `git log` to see examples.
Not `gallery: cache performance`, but `gallery: Increase cache performance`.

src/gallery.hpp
litoj: Perhaps no functional one, put practically the original method name described a different behaviour than the comment that describes the method. The method name suggested all thumbnails are cleared, when in reality only the ones that are not visible are. This rename makes the code more maintainable and easier to understand.

Please reconsider this. There is no reason to avoid every change that is not purely functional. That way you're basically preventing any improvements to the readability to the code and thereby increasing the barrier of entry for new people that might want to help the project as well.

Would `clear_unused_thumbnails` be better?

src/gallery.hpp
```diff
@@ -222,7 +222,11 @@ class Gallery : public AppMode {
         ImageEntryPtr entry;
         Pixmap pm;
     };
-    std::deque<ThumbEntry> cache;   ///< Thumbnails cache
+    /**
+     * Thumbnails cache.
+     * ImageEntry caching allows determining the current index of the thumbnail.
+     */
+    std::unordered_map<std::filesystem::path, ThumbEntry> cache;
```
artemsen: May be `std::map<ImageEntryPtr, ThumbEntry> cache` would be better. It strange to search by path when we have unique object.
Especially that unordered_map doesn't give any performance gain compared to deque, it's still O(n).

litoj: What operation are you refering to when you are comparing unordered_map and deque?

Adding is at most O(n), usually O(1) and everything else is always O(1) for unordered map. Deque doesn't have any benefits over a map, because it is also at worst O(n) for adding elements, but is always O(n) for finding one.

> May be std::map<ImageEntryPtr, ThumbEntry> cache would be better

I had it there originally, but then, when implementing the bulkchange (see
https://github.com/litoj/swayimg/blob/597a97615c7b4f2926ef7e8bef67d391a8407942/src/gallery.cpp#L387 )
I had to change it to use the path, because imagelist is also using path as the key and the ChangeTracker consists of sets of paths. Though it would be possible to change it to entries for sure at the cost of some extra copying of data in https://github.com/litoj/swayimg/blob/597a97615c7b4f2926ef7e8bef67d391a8407942/src/imagelist.cpp#L240 . On the other hand perhaps it's a worthy tradeoff to keep the key consistency as well as the return data of the public imagelist methods.

---

OK, half a day later and now I also have results of the profiling. I am surprised to say this, but both of our expectations were wrong.

The best approach in every way (except gallery window render for some reason) is the one already implemented in this PR - using the path as the index.

## Legend

1. `4b2a0c4` - lua: Load user's shared libs from config dir
2. `3b3351d` - gallery: path->thumbentry
3. `abcdefg` - gallery: imgentry->pixmap

## Total Times (s)

|  Commit | Best | Avg | Median | Worst |
| ------- | ---- | --- | ------ | ----- |
| 4b2a0c4 |  648 | 665 |    667 |   677 |
| 3b3351d |  644 | 651 |    652 |   660 |
| abcdefg |  646 | 665 |    669 |   675 |

## Call Averages (ms)

### APP

|  Commit | add |  load | remove | switch |
| ------- | --- | ----- | ------ | ------ |
| 4b2a0c4 | 1.2 | 119.5 |    1.2 |    1.4 |
| 3b3351d | 1.2 | 118.9 |    1.1 |    1.4 |
| abcdefg | 1.2 | 118.3 |    1.1 |    1.4 |

### GAL

|  Commit | clearthumb | getthumb | ilchange | loadthumb | redraw |
| ------- | ---------- | -------- | -------- | --------- | ------ |
| 4b2a0c4 |        0.0 |      0.0 |      0.5 |      51.3 |    6.9 |
| 3b3351d |        0.0 |      0.0 |      0.5 |      50.4 |    6.8 |
| abcdefg |        0.0 |      0.0 |      0.5 |      51.2 |    6.7 |

### LAY

|  Commit | select | update |
| ------- | ------ | ------ |
| 4b2a0c4 |    0.4 |    0.4 |
| 3b3351d |    0.4 |    0.3 |
| abcdefg |    0.4 |    0.4 |

I am not surprised a map is faster than a queue, but I am honestly surprised it is faster than indexing by a pointer. I thought it would be comparing just the literal pointer, but that would surely yield at least equivalently as fast loadthumb times, idk. 

Anyway, does this satisfy your curiosity for the benefits (though, yes, agreeably small) of this approach? 10% benefit in overal performance application performance is pretty good imo.

Here is the config file I made if you want to use it for stresstesting swayimg. It aims to make a workload similar to what searching through the gallery (and the possibility to filter files) could look like if implemented using the currently available options.

[asyncprofiler.patch.txt](https://github.com/user-attachments/files/28181751/asyncprofiler.patch.txt)
[perf_test.lua.txt](https://github.com/user-attachments/files/28181752/perf_test.lua.txt)

artemsen: > What operation are you refering to when you are comparing unordered_map and deque?

`clear_thumbnails` just changed the name, so I talked about `get_thumbnail`. Search.

> Adding is at most O(n), usually O(1) and everything else is always O(1) for unordered map. Deque doesn't have any benefits over a map, because it is also at worst O(n) for adding elements, but is always O(n) for finding one.

This is not entirely true. Search in `std::unordered_map` is `O(n)` (like in `deque`), in `std::map` - `O(log n)`.
Insertion and deletion operations of cache elements are quite rare compared to searches (which are used during redrawing), so search speed is more important. That's why I suggested to use map with `ImageEntryPtr` as key.

> Anyway, does this satisfy your curiosity for the benefits (though, yes, agreeably small) of this approach? 10% benefit in overal performance application performance is pretty good imo.

Yes, that satisfies me =)
Although it's not entirely clear how much data was tested (100 thumbnails or 10000). And the requirement to install the swi plugin to run the test doesn't inspire confidence.
It's far away from "10% benefit in overall application performance", but there is clearly progress.
I would prefer the fast redrawing option. Milliseconds during loading are not as important as milliseconds during rendering.

litoj: > Search in std::unordered_map is O(n) (like in deque), in std::map - O(log n).

Hash maps work the following way:
1. hashes the key (for a path that means reading all the chars in the path string, for a pointer it should be able to just use it as-is)
2. indexes with the hash (O(1)) into the [linkedlist of buckets](https://stackoverflow.com/a/31102576/12174842)
3. compares the key with each of the elements in the given bucket (O(n) for the contents of the bucket - usually <=2 because the map resizes at element count > approx. 1/2 bucket count to ensure minimum hashing colisions)

[SO benchmarked](https://stackoverflow.com/questions/36392583/is-an-unordered-map-really-faster-than-a-map-in-practice), [find is O(1)](https://stackoverflow.com/a/7019541/12174842)

> Although it's not entirely clear how much data was tested (100 thumbnails or 10000).

My bad. I ran it 10 times in a 1248x1536 window with the settings that are specified there at the beginning and on my gallery that has 2300 files in multiple recursive subfolders (so it tests the copying of added-file vectors in il:add_dir()) and a couple of video clips, so it also tests that they get removed when it tries to load their thumbnails.

> And the requirement to install the swi plugin to run the test doesn't inspire confidence.

Wrong assumption, just run it as your config :).

> I would prefer the fast redrawing option. Milliseconds during loading are not as important as milliseconds during rendering.

Done, I've added also a change that improves the redraw slightly furter by saving the selected index when it comes across it during the first loop, so that it can just index it directly after the loop instead of going through the scheme all again to find just the selected item. I have also made the `draw()` method take the selection as a boolean instead of querying it on its own repeatedly for each image. Since we already know the value when we call it there is no necessity to recalculate it. This also allows the future possibility to add multiple-selection etc.

---

PS: I know I have added another unrelated commit. But given that this review is pretty much done, is it really much of a problem if I fix a non-logical issue?

artemsen: > Hash maps work the following way

I know how a hash map works. That's why I suggest storing ImageEntriyPtr as keys, not hashes.

image

src/gallery.cpp
```diff
@@ -404,11 +397,11 @@ void Gallery::handle_imagelist(const ImageListEvent event,
     Application::redraw();
 }
 
-void Gallery::draw(const Layout::Thumbnail& tlay, Pixmap& wnd)
+void Gallery::draw(const Layout::Thumbnail& tlay, Pixmap& wnd,
+                   const bool selected)
```
artemsen: You saved a few CPU instructions, but you also duplicated the parameters and added the potential to call draw a selected thumbnail when it isn't.
In other words, you can do `draw(NonSelectedThumb, pm, true)`.
I would prefer an unambiguous API to such a minor time savings.

src/gallery.hpp
```diff
@@ -217,12 +218,11 @@ class Gallery : public AppMode {
...
-    struct ThumbEntry {
-        ImageEntryPtr entry;
-        Pixmap pm;
-    };
-    std::deque<ThumbEntry> cache;   ///< Thumbnails cache
+    /**
+     * Thumbnails cache.
+     * ImageEntry caching allows determining the current index of the thumbnail.
+     */
+    std::unordered_map<ImageEntryPtr, Pixmap> cache;
```
artemsen, line 224: You marked our discussion as resolved, which is probably why you missed my last comment.
Let's check how the `map` will behave.

litoj: Indeed I have missed it, thank you for pointing out. However, I do not know what do you mean by "check how the map will behave". I've already provided the benchmark results, which seem quite conclusive IMO (I've improved the calculation precission slightly in the results in the other PR, which include this change as a baseline comparison next to the unchanged version) and my previous comment also included sources of benchmarks comparing `std::map` and `std::unordered_map`.

> I know how a hash map works. That's why I suggest storing ImageEntriyPtr as keys, not hashes.

And you were absolutely right on that one. But if you know how it works, why do you then want to keep checking performance when you know its O(1) in practise?

Yes, theoretically the worst time of a map is O(n) because the values can all get hashed to the same number, in which case the entire map will be in one bucket (ie linkedlist/`std::list`), but that is as unlikely as that the sorting that is used everywhere will be O(n²), because QuickSort is O(nlogn) only on average :D.

Also, regarding things that may have been missed, do you still want me to undo the rename of `clear_thumbnails` or can it stay renamed to the more representative `clear_unused_thumbnails`?

_PS: the draw() change is ready, I will push it together with the next changes if you will require any._

artemsen: > theoretically the worst time of a map is O(n)

Okay, you've convinced me, let's stop at unordered_map =)

> Also, regarding things that may have been missed, do you still want me to undo the rename of `clear_thumbnails` or can it stay renamed to the more representative `clear_unused_thumbnails`?

I wouldn't bother with this at all, but if you want to rename it, it's better to use the term `visible`, just like the variables in the code of this function. May be `clear_invisible_thumbnails`?

PR #490, opened 2026-05-20
List optimizations

litoj:
2nd change in a series of changes making large imagelist changes more performant

relates to https://github.com/artemsen/swayimg/issues/481

Tracking the newly added images is done via a inserter function in order to prepare for the bulk update which will be using an unordered set instead of a vector.

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -162,7 +162,7 @@ ImageEntryPtr Application::add_image_entry(const std::filesystem::path& path)
     // update image list text info
     Text& text = Text::self();
     text.set_field(Text::FIELD_LIST_INDEX,
-                   std::to_string(current_mode()->current_entry()->index));
+                   std::to_string(current_mode()->current_entry()->index + 1));
     text.set_field(Text::FIELD_LIST_TOTAL, std::to_string(il.size()));
     text.update();
 
@@ -185,7 +185,7 @@ void Application::remove_image_entry(const std::filesystem::path& path)
     // update image list text info
     Text& text = Text::self();
     text.set_field(Text::FIELD_LIST_INDEX,
-                   std::to_string(current_mode()->current_entry()->index));
+                   std::to_string(current_mode()->current_entry()->index + 1));
     text.set_field(Text::FIELD_LIST_TOTAL, std::to_string(il.size()));
     text.update();
 
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -151,7 +151,7 @@ ImageEntryPtr Application::add_image_entry(const std::filesystem::path& path)
 {
     ImageList& il = ImageList::self();
 
-    const std::vector<ImageEntryPtr> entries = il.add(path);
+    const std::list<ImageEntryPtr> entries = il.add(path);
     if (entries.empty()) {
         return nullptr;
     }
--- a/src/image.hpp
+++ b/src/image.hpp
@@ -35,6 +35,13 @@ struct ImageEntry {
     static constexpr const char* SRC_STDIN = "stdin://";
     // Special prefix used to load images from external command output
     static constexpr const char* SRC_EXEC = "exec://";
+
+    /**
+     * Check if path is a special source.
+     * @param path path to check for special source format
+     * @return true if path starts with stdin:// or exec://
+     */
+    static bool is_special(const std::string& path);
 };
 
 using ImageEntryPtr = std::shared_ptr<ImageEntry>;
--- a/src/image.hpp
+++ b/src/image.hpp
@@ -17,7 +17,7 @@ struct ImageEntry {
     std::filesystem::path path; ///< Path to the image file
     std::time_t mtime;          ///< File modification time
     size_t size;                ///< Size of the image file
-    size_t index;               ///< Image index in image list
+    size_t index;               ///< Image index in image list (0-based)
     bool mark;                  ///< Marked image flag
 
     /**
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -158,7 +158,7 @@ ImageEntryPtr ImageList::load(const std::vector<std::filesystem::path>& sources)
     // files to the list
     adjacent = false;
 
-    // get first entry
+    // get first entry of the source list (not of the contents of the sources)
     ImageEntryPtr first = nullptr;
     for (auto& path : sources) {
         first = find(path);
@@ -179,7 +179,7 @@ ImageEntryPtr ImageList::load(const std::vector<std::filesystem::path>& sources)
             }
         }
     }
-    return first ? first : *entries.begin();
+    return first ? first : entries.front();
 }
 
 ImageEntryPtr ImageList::load(const std::filesystem::path& list_file)
@@ -203,13 +203,10 @@ ImageEntryPtr ImageList::load(const std::filesystem::path& list_file)
     return load(sources);
 }
 
-std::vector<ImageEntryPtr> ImageList::add(const std::filesystem::path& path)
+std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path)
 {
     const std::unique_lock lock(mutex);
-
-    std::vector<ImageEntryPtr> entries = add(path, true);
-
-    return entries;
+    return add(path, true);
 }
 
 ImageEntryPtr ImageList::remove(const ImageEntryPtr& entry, bool forward)
@@ -392,11 +389,9 @@ ssize_t ImageList::distance(const ImageEntryPtr& from, const ImageEntryPtr& to)
     return static_cast<ssize_t>(to->index) - static_cast<ssize_t>(from->index);
 }
 
-std::vector<ImageEntryPtr> ImageList::add(const std::filesystem::path& path,
-                                          const bool ordered)
+std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path,
+                                        const bool ordered)
 {
-    std::vector<ImageEntryPtr> entries;
-
     if (path.string().starts_with(ImageEntry::SRC_STDIN) ||
         path.string().starts_with(ImageEntry::SRC_EXEC)) {
         ImageEntryPtr entry = std::make_shared<ImageEntry>();
@@ -405,39 +400,47 @@ std::vector<ImageEntryPtr> ImageList::add(const std::filesystem::path& path,
         entry->size = 0;
         entry->index = 0;
         if (add_entry(entry, ordered)) {
-            entries.push_back(entry);
+            return { entry };
         }
-    } else {
-        std::filesystem::path abs_path;
-        try {
-            abs_path = std::filesystem::absolute(path).lexically_normal();
-        } catch (...) {
-            Log::warning("Invalid path {}, skipped", path.string());
-            return {};
-        }
-        if (!std::filesystem::exists(path)) {
-            Log::warning("File {} not found, skipped", path.string());
-            return {};
-        }
+        return {};
+    }
 
-        if (std::filesystem::is_directory(path)) {
-            entries = add_dir(abs_path, ordered);
-        } else if (adjacent) {
-            const std::vector<ImageEntryPtr> edir =
-                add_dir(abs_path.parent_path(), ordered);
-            entries.insert(entries.end(), edir.begin(), edir.end());
-        } else {
-            entries.push_back(add_file(abs_path, ordered));
+    std::filesystem::path abs_path;
+    try {
+        abs_path = std::filesystem::absolute(path).lexically_normal();
+    } catch (...) {
+        Log::warning("Invalid path {}, skipped", path.string());
+        return {};
+    }
+    if (!std::filesystem::exists(path)) {
+        Log::warning("File {} not found, skipped", path.string());
+        return {};
+    }
+
+    if (std::filesystem::is_directory(path)) {
+        auto ret = add_dir(abs_path);
+        if (!ret.empty() && ordered) {
+            sort(false);
+        }
+        return ret;
+    } else if (adjacent) {
+        auto ret = add_dir(abs_path.parent_path());
+        if (!ret.empty() && ordered) {
+            sort(false);
+        }
+        return ret;
+    } else {
+        if (const ImageEntryPtr& entry = add_file(abs_path, ordered)) {
+            return { entry };
         }
     }
 
-    return entries;
+    return {};
 }
 
-std::vector<ImageEntryPtr> ImageList::add_dir(const std::filesystem::path& path,
-                                              const bool ordered)
+std::list<ImageEntryPtr> ImageList::add_dir(const std::filesystem::path& path)
 {
-    std::vector<ImageEntryPtr> entries;
+    std::list<ImageEntryPtr> added;
 
     if (fsmon) {
         FsMonitor::self().add(path);
@@ -448,24 +451,19 @@ std::vector<ImageEntryPtr> ImageList::add_dir(const std::filesystem::path& path,
             const std::filesystem::path& sub_path = it.path();
             if (std::filesystem::is_directory(sub_path)) {
                 if (recursive) {
-                    std::vector<ImageEntryPtr> added = add_dir(sub_path, false);
-                    entries.insert(entries.end(), added.begin(), added.end());
+                    added.splice(added.end(), add_dir(sub_path));
                 }
             } else {
                 const ImageEntryPtr entry = add_file(sub_path, false);
                 if (entry) {
-                    entries.push_back(entry);
+                    added.emplace_back(entry);
                 }
             }
         }
     } catch (const std::filesystem::filesystem_error&) {
     }
 
-    if (!entries.empty() && ordered) {
-        sort(false);
-    }
-
-    return entries;
+    return added;
 }
 
 ImageEntryPtr ImageList::add_file(const std::filesystem::path& path,
@@ -533,7 +531,7 @@ bool ImageList::add_entry(ImageEntryPtr& entry, const bool ordered)
     return true;
 }
 
-void ImageList::sort(bool locked)
+void ImageList::sort(const bool locked)
 {
     if (locked) {
         mutex.lock();
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -256,8 +256,7 @@ ImageEntryPtr ImageList::find(const std::filesystem::path& path)
     if (search.empty()) {
         return nullptr;
     }
-    if (!search.starts_with(ImageEntry::SRC_STDIN) &&
-        !search.starts_with(ImageEntry::SRC_EXEC)) {
+    if (!ImageEntry::is_special(search)) {
         try {
             search = std::filesystem::absolute(path).lexically_normal();
         } catch (...) {
@@ -392,14 +391,8 @@ ssize_t ImageList::distance(const ImageEntryPtr& from, const ImageEntryPtr& to)
 std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path,
                                         const bool ordered)
 {
-    if (path.string().starts_with(ImageEntry::SRC_STDIN) ||
-        path.string().starts_with(ImageEntry::SRC_EXEC)) {
-        ImageEntryPtr entry = std::make_shared<ImageEntry>();
-        entry->path = path;
-        entry->mtime = 0;
-        entry->size = 0;
-        entry->index = 0;
-        if (add_entry(entry, ordered)) {
+    if (ImageEntry::is_special(path)) {
+        if (const ImageEntryPtr& entry = add_special_source(path, ordered)) {
             return { entry };
         }
         return {};
@@ -433,9 +426,8 @@ std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path,
         if (const ImageEntryPtr& entry = add_file(abs_path, ordered)) {
             return { entry };
         }
+        return {};
     }
-
-    return {};
 }
 
 std::list<ImageEntryPtr> ImageList::add_dir(const std::filesystem::path& path)
@@ -495,6 +487,22 @@ ImageEntryPtr ImageList::add_file(const std::filesystem::path& path,
         FsMonitor::self().add(path);
     }
     return entry;
+}
+
+ImageEntryPtr ImageList::add_special_source(const std::filesystem::path& path,
+                                            const bool ordered)
+{
+    assert(is_special_source(path));
+
+    ImageEntryPtr entry = std::make_shared<ImageEntry>();
+    entry->path = path;
+    entry->mtime = 0;
+    entry->size = 0;
+    entry->index = 0;
+    if (add_entry(entry, ordered)) {
+        return entry;
+    }
+    return nullptr;
 }
 
 bool ImageList::add_entry(ImageEntryPtr& entry, const bool ordered)
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -209,7 +209,7 @@ std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path)
     return add(path, true);
 }
 
-ImageEntryPtr ImageList::remove(const ImageEntryPtr& entry, bool forward)
+ImageEntryPtr ImageList::remove(const ImageEntryPtr& entry, const bool forward)
 {
     assert(entry);
 
@@ -505,7 +505,7 @@ ImageEntryPtr ImageList::add_special_source(const std::filesystem::path& path,
     return nullptr;
 }
 
-bool ImageList::add_entry(ImageEntryPtr& entry, const bool ordered)
+bool ImageList::add_entry(const ImageEntryPtr& entry, const bool ordered)
 {
     auto [it, inserted] = entry_map.insert({ entry->path, entry });
 
@@ -562,18 +562,17 @@ void ImageList::sort(const bool locked)
                   });
     }
 
-    reindex(0);
+    reindex();
 
     if (locked) {
         mutex.unlock();
     }
 }
 
-void ImageList::reindex(size_t index)
+void ImageList::reindex(const size_t index)
 {
-    auto end = entries.end();
-    auto it = entries.begin() + index;
-    while (it != end) {
-        (*it++)->index = index++;
+    // i-- to reindex also at the actual index
+    for (size_t i = entries.size(); i-- > index;) {
+        entries[i]->index = i;
     }
 }
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -208,9 +208,6 @@ std::vector<ImageEntryPtr> ImageList::add(const std::filesystem::path& path)
     const std::unique_lock lock(mutex);
 
     std::vector<ImageEntryPtr> entries = add(path, true);
-    if (!entries.empty()) {
-        reindex();
-    }
 
     return entries;
 }
@@ -227,9 +224,9 @@ ImageEntryPtr ImageList::remove(const ImageEntryPtr& entry, bool forward)
 
     entry_map.erase(entry->path);
     entries.erase(entries.begin() + entry->index);
-    entry->remove();
+    reindex(entry->index);
 
-    reindex();
+    entry->remove();
 
     return next;
 }
@@ -341,7 +338,6 @@ ImageEntryPtr ImageList::get(const ImageEntryPtr& from, const Dir dir)
     }
 
     auto it = entries.begin() + index;
-    assert(it != entries.end());
 
     const std::filesystem::path from_parent = from->path.parent_path();
     if (dir == Dir::NextParent) {
@@ -529,7 +525,9 @@ bool ImageList::add_entry(ImageEntryPtr& entry, const bool ordered)
                               });
         }
 
+        entry->index = it == entries.end() ? entries.size() : (*it)->index;
         entries.insert(it, entry);
+        reindex(entry->index + 1); // reindex all entries after this one
     }
 
     return true;
@@ -558,17 +556,18 @@ void ImageList::sort(bool locked)
                   });
     }
 
-    reindex();
+    reindex(0);
 
     if (locked) {
         mutex.unlock();
     }
 }
 
-void ImageList::reindex()
+void ImageList::reindex(size_t index)
 {
-    size_t index = 0;
-    for (auto& it : entries) {
-        it->index = index++;
+    auto end = entries.end();
+    auto it = entries.begin() + index;
+    while (it != end) {
+        (*it++)->index = index++;
     }
 }
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -225,9 +225,8 @@ ImageEntryPtr ImageList::remove(const ImageEntryPtr& entry, bool forward)
 
     const std::unique_lock lock(mutex);
 
+    entry_map.erase(entry->path);
     entries.erase(entries.begin() + entry->index);
-    duplicates.erase(entry->path);
-
     entry->remove();
 
     reindex();
@@ -274,11 +273,8 @@ ImageEntryPtr ImageList::find(const std::filesystem::path& path)
 
     const std::shared_lock lock(mutex);
 
-    auto it = std::find_if(entries.begin(), entries.end(),
-                           [&search](const ImageEntryPtr& entry) {
-                               return search == entry->path;
-                           });
-    return it == entries.end() ? nullptr : *it;
+    auto it = entry_map.find(search);
+    return it == entry_map.end() ? nullptr : it->second;
 }
 
 std::vector<ImageEntry> ImageList::get_all()
@@ -509,10 +505,11 @@ ImageEntryPtr ImageList::add_file(const std::filesystem::path& path,
 
 bool ImageList::add_entry(ImageEntryPtr& entry, const bool ordered)
 {
-    if (duplicates.contains(entry->path)) {
+    auto [it, inserted] = entry_map.insert({ entry->path, entry });
+
+    if (!inserted) {
         return false; // already exists
     }
-    duplicates.insert(entry->path);
 
     if (!ordered || entries.empty() || order == Order::None) {
         entry->index = entries.size();
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -343,20 +343,19 @@ ImageEntryPtr ImageList::get(const ImageEntryPtr& from, const Dir dir)
                 return *it;
             }
         }
-    } else if (dir == Dir::PrevParent && index > 0) {
+        return nullptr;
+    }
+    if (dir == Dir::PrevParent) {
         const auto start = entries.begin();
-        while (true) {
-            it--;
+        while (it-- != start) {
             if (from_parent != (*it)->path.parent_path()) {
                 return *it;
             }
-
-            if (it == start) {
-                break;
-            }
         }
+        return nullptr;
     }
 
+    assert(false && "unhandled iterating position");
     return nullptr;
 }
 
@@ -392,7 +391,8 @@ std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path,
                                         const bool ordered)
 {
     if (ImageEntry::is_special(path)) {
-        if (const ImageEntryPtr& entry = add_special_source(path, ordered)) {
+        const ImageEntryPtr entry = add_special_source(path, ordered);
+        if (entry) {
             return { entry };
         }
         return {};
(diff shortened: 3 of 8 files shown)


artemsen:
AI again. Okay then.
Please add a description to each commit: why it's being done (not "how," but "why"), why the changed code was bad and why it's now good, how performance measurements were taken, and what results they yielded.

litoj:
> AI again.

How did you know? :D, but no, actually, only the voiditerator was. It would be much more helpful to hear what you dislike about the code (unless what you've said is all that's wrong with it, but then it would be just git and not the code, so you'd have no reason to call it AI 🤔).

I have spent a whole day trying to debug the existing code that was not meant for quick list changes and I have no doubts there is something that could have been done differently or better, but the original code was so inefficient that this is still an improvement in terms of both simplicity as well as performance.

I have intentionally split the changes to commits of individual smaller changes instead of one big change I had originally to make it easier to review and each commit says perfectly all it does with the goal being the same everywhere - increase performance / fix bugs. Of course, I will gladly add more description to the commits themselves, but all the places that needed explaining already say what's the purpose of the code right along the code itself (with the general goal being performance, if not described otherwise).

My performance metrics are just practical observations atm - logically, if you replace a linear find() with an instant index lookup, you will have sped up the code. Yes, I understand vector required allocating the whole space for each shared_ptr of the image entries ahead of time and can resize, but that happens way fewer times than having to search through the entire list every time we want to do anything with it. I assume you already know that, so I'd rather ask what would you like to know the performane (other than the first commit :)) / meaning of that is not obvious to you already?

meson.build
```diff
@@ -30,6 +30,21 @@ if version == '0.0.0'
     endif
 endif
 
+# ensuer memory errors print proper stacktrace when in debug build
+if get_option('buildtype').startswith('debug')
```
artemsen: I'm glad you discovered sanitizers; I always build the project with similar parameters, but...

- With this patch you have hard-linked the project to GCC. Other compilers may not support these flags. Just for example: building a project using clang with address sanitizer requires adding `-fopenmp=libgomp`, otherwise the project will not link (because of libraw).
- Now it is impossible to use another sanitizer, because in most cases they do not work simultaneously. For example, if I need `-fsanitize=thread`, I will have to revert this patch.
- There is a standard `NDEBUG` flag that used in asserts and handles by meson. You don't need to create new `DEBUG` macro.
- You put different tasks in one PR (again). This change doesn't optimize image list.

src/imagelist.hpp
```diff
@@ -176,15 +189,16 @@ class ImageList {
...
      * Reindex the image list.
+     * @param index (0-based) to start reindexing from
      */
-    void reindex();
+    void reindex(size_t index);
 
 private:
-    std::list<ImageEntryPtr> entries; ///< List of image entries
-    std::shared_mutex mutex;          ///< Image list mutex
+    std::vector<ImageEntryPtr> entries; ///< List of image entries
```
artemsen, line 189: In version 4.0, I converted the image list from array to real list. This made many things easier and faster, and most importantly, it got rid of the problems with the index, which you are now trying to get back.
Getting a record by index from an array is indeed a very fast operation, but it is not free and has to be paid for elsewhere, which negates all this optimization.
I can't remember all the details, but the main problem is the indexes themselves. If they are constants, then we have holes in the array (this is how version 1.0 worked). If they are modified at runtime when records are added/removed, then they cannot be accessed from different threads without global list lock.
The second problem is the slow insertion and deletion of records. This isn't as relevant now, as there's not much data (only shared pointers) but it's still an issue.

Working with the image list hides many pitfalls in a multithreaded environment, and there's a high risk that such a transition will break things rather than improve them. Such a transition requires careful manual verification, and I have neither the time nor, more importantly, the desire to do this. Especially since there are no optimization metrics, so it is impossible to assess how much faster the new implementation is than the existing one (or whether it is faster at all).

Taking all this into account, I am not ready to accept these changes.

src/imagelist.cpp
```diff
@@ -337,16 +329,17 @@ ImageEntryPtr ImageList::get(const ImageEntryPtr& from, const Dir dir)
         return it == entries.end() ? entries.front() : *it;
     }
 
-    auto it = std::find(entries.begin(), entries.end(), from);
+    // get up-to-date value
+    const size_t index = entry_map[from->path]->index - 1;
 
     if (dir == Dir::Next) {
-        return ++it == entries.end() ? nullptr : *it;
+        return index == entries.size() - 1 ? nullptr : entries[index + 1];
```
artemsen, line 330: 1-based is a bad idea. It is not Lua, arrays in C starts from 0.
Older versions of Swayimg uses the same logic with 1-based index and I often got confused and made "off-by-one" errors.
So in 5.0 I changed it to 0-based to remove all these +1/-1.

src/pixmap.cpp
```diff
@@ -322,7 +322,8 @@ void Pixmap::rectangle(const Rectangle& rect, const argb_t& color,
 void Pixmap::mask(const Pixmap& pm, const Point& pos, const argb_t& color)
 {
     assert(format() == Format::RGB || format() == Format::ARGB);
-    assert(pm.format() == Format::GS);
+    // Not actually always true
+    // assert(pm.format() == Format::GS);
```
artemsen: It _must_ be grayscale. It's bug if not, mask can not be RGB, it's a 8-bit alpha channel.

artemsen (commented):
I only looked at the first two commits. Since your PR is based on using vector instead of a list, which I fundamentally disagree with, I didn't review the remaining commits.

> Sometimes it feels like you're trying to insult everyone who has less experience and different approaches to coding than you do

No, that's not true. But in some ways you are right - I am indeed biased towards your PRs. There are a number of reasons for this:
1. Your PRs often contain a lot of unnecessary changes, and extra code is an opportunity to make more mistakes. Not to mention that in one PR you have changes for completely different tasks.
2. Some of your decisions are based solely on mathematics, without taking into account the specifics of using structures in Swayimg. This often makes the code "more correct," but does not provide any real improvements, turning the changes into cosmetic ones.
3. You're performing minimal testing, apparently relying on automated tests that don't cover even 10% of all cases. This leads to disastrous results, and fundamental changes like those in this PR require very thorough testing due to potential multithreading issues.
4. The PRs usually (except RFC) imply that you've solved some problem and want to merge the code into the master. It's ready and tested. But you create a PR that initially doesn't even pass CI, even though you have the opportunity to test it on your fork. As a result, you're fixing it on the fly, I get dozens of "litoj pushed 1 commit" emails, but I don't even know if the PR is ready for review or if you're still in the process. Making edits to a PR is fine, but you should do them based on the reviewer's comments, not change everything on the fly, because that's more convenient.

Therefore, the priority of analyzing your code shifts from "making it better" to "not breaking what's already there".
(Perhaps I wrote too bluntly, but I certainly did not intend to humiliate or insult you.)

src/imagelist.cpp
litoj: Yes, that is why I save the `entry->index` value first to deduct the 1 from it to make it 0-based and use it like a proper index here, too.

What exactly do you refer to as 1-based here? `entries[index+1]` is just the next entry index.

artemsen: I use 1-based indexing because it's just a number to display in the text block (showing 0 there would be weird).
You use an index to access array elements, but for some reason you also left it as 1-indexed.
Why not use 0-based then and get rid of all these +1/-1, and use +1 only for the text block? 
However, this is no longer important.

litoj: I did _not_ leave it 1-based. How are you coming to that conlusion when all the comparisons are clearly for a 0-based index as is the line that creates the variable? :thinking: 

Btw I am working on the performance benchmark as you wished, so please don't disregard the entire PR yet. I understand your points, but you are basing most of them on the assumption that I did no testing, which is the opposite of truth, because there is an entire commit just for fixing all the concurrency issues that I discovered and spent a whole day on debugging, as I described in the initial answer already.

Also sorry for the commit spam previously. I thought it sends emails only for comments. My clangd also doesn't seem to be using the same clang-tidy config as the CI despite using the compile commands for `meson -Dtests=enabled`. So I don't see the errors that pop up in the CI here.

I will try to use the branch CI on a separate branch before comitting next time.

artemsen: > I did _not_ leave it 1-based. How are you coming to that conlusion when all the comparisons are clearly for a 0-based index as is the line that creates the variable? 🤔

image

> don't disregard the entire PR yet

Ok, but please go over the code again, you sometimes delete pieces of code, breaking existing functionality (for example getting the first entry in ImageList::load).

> My clangd also doesn't seem to be using the same clang-tidy config as the CI despite using the compile commands for meson -Dtests=enabled

clang-tidy doesn't depends on tests, it uses .clang-tidy in the project's root.

litoj: > clang-tidy doesn't depends on tests, it uses .clang-tidy in the project's root.

In that case I must have something wrong with my clangd because even with `clangd --clang-tidy` I don't see the `could be const` errors that the CI discoveres :/.

The screenshot you provided is of the reindex function. I now take it you had issue with the value in the image entry `index` field being left as a 1-based number as it was originally. 

Since the line that you marked the comment on doesn't use `entry->index` I thought you were talking about the `size_t index = entry->index-1` variable, which is converted to a regular index as it should be. I now take it that you meant the entry and not the variable, right?

We could have it 0-based and convert to 1-based at the lua table conversion level. Would that work for you?

> (for example getting the first entry in ImageList::load).

That was my latest added change. I can drop that commit if you wish, but I feel that the first entry should be the entry that ends up first in the sorted imagelist - not just the first loaded file, so that what we return corresponds to the sorting order the user has chosen. That way when the added directory is opened in viewer mode the file that is displayed is always the first file of the directory from the user's POV as well, as it should be. Wouldn't you agree?

artemsen: > even with `clangd --clang-tidy`

You can take a look at how [CI does it](https://github.com/artemsen/swayimg/blob/7c491d4c716fa4dde614e8be58d6cb308802da48/.github/workflows/Arch.yml#L55).
`ninja -C MY_BUILD_PATH clang-tidy`

> that you meant the entry and not the variable, right?

Yes, it's entry index is 1-base.

> We could have it 0-based and convert to 1-based at the lua table conversion level. Would that work for you?

Yes, but not in Lua, it should be incremented only for UI in the Text class.
But these are trifles, I wouldn’t bother with this until I’m confident in using the array.

I had an idea to make not an array or a list, but a set. This would provide fast insertion/deletion and searching, but would require implementing a comparison operator tied to the sort order. But I never got around to experimenting with it.

litoj: > I had an idea to make not an array or a list, but a set. This would provide fast insertion/deletion and searching, but would require implementing a comparison operator tied to the sort order.

My two cents on this interesting idea:

An ordered `set` would be able to replace both a map and a list. However, searching would _not_ be faster than an unordered map I am suggesting (set=O(logn) vs umap=O(1)). So it would negatively impact the most common (get) operation.

That means it's _not suited for the usecase of swayimg_, but let's dig deeper and see what benefits there are to be had.

- Inserting and deleting could be O(logn) instead of O(n), but only if we remove `entry->index` field entirely so that reindexing could be avoided as well as sorting (O(n) + O(nlogn)) and replaced it with a call to sth like a `il:indexof(ieptr)` for whenever the index value is needed.
- That method would get the iterator for the entry (O(logn)) and return the difference from the beginning iterator.
- However, from what I've found online the `std::distance()` works both ways only for vectors/lists and is O(n) for sets.
- In fact the proposed solution to finding the distance in a set is [to use a sorted vector](https://stackoverflow.com/a/12533895/12174842) like we are already doing with `entries`. Meaning there is likely no more efficient solution than the one I am proposing here combined with the updating index as it is currently :sweat_smile: 

Regarding the comparator problem:

my idea, which I want to implement since I first saw the current comparator implementation, is to construct a comparator lambda - a custom one for each combination of parameters (4 x 2 for reverse + random -> 9 + I'd add an option to use a comparator passed in from lua) to make the comparing a simple process instead of checking the current settings for every comparison that occurs, which is the current approach. It would speed up sorting a little, so a faster startup time (always nice). I am definitely going to make a PR with it at some point this year.

---

> In version 4.0, I converted the image list from array to real list. This made many things easier and faster, and most importantly, it got rid of the problems with the index, which you are now trying to get back.

I assume the code for 4.0 was much, _much_ different, since the only negative impact of using an array in the current version is that during initial load the vector will have to resize multiple times, in turn driving up the load time slightly. However, everything else is multiple times faster. Even though the load time increases by 16%, the rendering speed increases by at least as much. Surely this is a worthy sacrafice ;)

The following are my benchmark results for the first 4 commits that aim to improve performance. In summary, the vector cuts through everything except load time, and map fixes the `find` speed. Better reindexing improves add and remove time around a margin-of-error amount and the lambda for collecting added files does nothing (probably would be more impactful if we were passing a reference to the main vector, but then we couldn't collect into a set the same way, which could come handy in the future `il:set_entries`).

I also included the best gallery performer that is in the other PR to give a better impactfulness-reference point.

## Legend

1. `4b2a0c4` - lua: Load user's shared libs from config dir
2. `7a9cdeb` - gallery: Avoid double-looping scheme during redraw
3. `79c5ac0` - imagelist: Make order lookups O(1) + 0-based entry index
4. `06e1d4b` - imagelist: Improve find performance by using unordered_map
5. `9aada0f` - imagelist: Optimize reindexing to skip unchanged entries
6. `1fc3fc5` - imagelist: Avoid copying lists of added items

## Total Cpu time (s)

|  Commit | Best | Avg | Median | Worst |
| ------- | ---- | --- | ------ | ----- |
| 4b2a0c4 |  641 | 661 |    667 |   676 |
| 7a9cdeb |  654 | 669 |    672 |   684 |
| 79c5ac0 |  610 | 623 |    625 |   640 |
| 06e1d4b |  606 | 614 |    611 |   629 |
| 9aada0f |  608 | 614 |    615 |   622 |
| 1fc3fc5 |  604 | 615 |    617 |   625 |

## Call duration avg (ms) per avg number of calls

Represents the average time the executions of the function accumulated over a single run of the benchmark. Headers specify the average number of calls of that function.

### APP

|  Commit | add<br>10000 | load<br>1.0 | remove<br>8100 | switch<br>5800 |
| ------- | ----- | ---- | ------ | ------ |
| 4b2a0c4 | 12300 |  122 |   9360 |   8270 |
| 7a9cdeb | 12400 |  123 |   9320 |   8240 |
| 79c5ac0 |  8590 |  142 |   6790 |   8130 |
| 06e1d4b |  8540 |  142 |   5030 |   8240 |
| 9aada0f |  8460 |  138 |   4970 |   8150 |
| 1fc3fc5 |  8480 |  138 |   4940 |   8150 |

### GAL

|  Commit | clearT<br>4000 | getT<br>340000 | ilchange<br>18000 | loadT<br>3400 | loadTs<br>4000 | redraw<br>4000 |
| ------- | ------ | ------ | -------- | ------ | ------ | ------ |
| 4b2a0c4 |    0.2 |    478 |     8680 | 372000 |   2100 |  28100 |
| 7a9cdeb |   20.8 |    135 |     8690 | 371000 |   1900 |  26800 |
| 79c5ac0 |   0.99 |    472 |     3020 | 367000 |    324 |  24000 |
| 06e1d4b |    0.5 |    452 |     3060 | 367000 |    321 |  23700 |
| 9aada0f |   1.39 |    453 |     3020 | 369000 |    322 |  23900 |
| 1fc3fc5 |   0.89 |    469 |     2990 | 368000 |    320 |  23800 |

### IL

|  Commit | add<br>10000 | dist<br>83000 | find<br>8100 | get<br>1400000 | get_dist<br>62000 | remove<br>7600 |
| ------- | ----- | ----- | ---- | ------- | -------- | ------ |
| 4b2a0c4 |  6570 |  2240 | 1970 |    9260 |      660 |    334 |
| 7a9cdeb |  6540 |  2240 | 1960 |    9200 |      660 |    335 |
| 79c5ac0 |  6420 |   5.4 | 1850 |     192 |     11.4 |   51.6 |
| 06e1d4b |  6350 |  5.09 |   28 |     201 |       12 |   52.3 |
| 9aada0f |  6310 |  5.99 | 28.1 |     194 |     11.4 |   43.3 |
| 1fc3fc5 |  6340 |  5.79 | 26.2 |     196 |     11.3 |   42.4 |

### LAY

|  Commit | select<br>5800 | update<br>28000 |
| ------- | ------ | ------ |
| 4b2a0c4 |   2240 |   9990 |
| 7a9cdeb |   2240 |  10000 |
| 79c5ac0 |    136 |    287 |
| 06e1d4b |    144 |    284 |
| 9aada0f |    137 |    285 |
| 1fc3fc5 |    138 |    284 |


I did some cleanup and removed unnecessary changes. Can we move forward with the current approach?

PS: I will push the single clang-tidy fix and remove the `src/compositor.cpp` change when there are more of them to be done or the PR is declared acceptable so that I don't spam with commits. And I also hate github for how it duplicates review comments to not be possible to be just answers to someone elses review... sorry for the mess.

litoj (commented):
Made some changes, hopefully the code is now clearer and the changes justified.

src/imagelist.cpp
```diff
@@ -147,40 +158,17 @@ ImageEntryPtr ImageList::load(const std::vector<std::filesystem::path>& sources)
...
-    mutex.unlock();
-
-    sort(true);
-
+    sort(false);
     // disable loading of adjacent files, otherwise fs mon will add unnecessary
     // files to the list
     adjacent = false;
 
-    // get first entry
```
artemsen, line 162: I suspect you didn't bother to figure out what this piece of code was for and just threw it away, right? =)

src/imagelist.hpp
```diff
@@ -60,9 +58,12 @@ class ImageList {
...
      * Add file or special source to the list.
      * @param path path to the file or special source
-     * @return array of added entries
+     * @param on_added callback invoked for each added entry
+     * @return first entry from source list
      */
-    std::vector<ImageEntryPtr> add(const std::filesystem::path& path);
+    ImageEntryPtr
+    add(const std::filesystem::path& path,
+        const std::function<void(const ImageEntryPtr&)>& on_added);
```
artemsen: I don't like this idea with callback.
Callback that call other callback that call other... Too complex.

src/imagelist.cpp
litoj: No, actually. My reasons for removing that code were two-fold.

The most obvious is consistency - your approach can open up the middle file of the list that the user sees, making it very confusing. Second one is reusability.

Both have the same solution - make the way the first entry is returned the same way everywhere where it may be required.

I understand it would be logical to sort files by their directory level and then name, but it should be implemented as a sorting comparator for the actual entry list, not just the returned file.

When this is implemented as a general comparator, it will be much more useful, because that is the order the user probably expects to see the images anyway.

I can implement it as a separate PR and replace it there.

artemsen: When I type `swayimg photo123.jpg` I expect photo123.jpg will be displayed. But with `true` in `swayimg.imagelist.enable_adjacent` the program will open the first image file in the current dir.

When I type `swayimg photo123.jpg ../dir` I expect photo123.jpg will be displayed even if there is a dir at second arg.

src/imagelist.hpp
litoj: there is only a single callback wrap, but that can be easily removed. It has two benefits - not having to create a list if you're not gonna use it (`load()`) and allowing immediate conversion to the structure in which you want to store it - vector/list/set/map - without having to copy all the data again.

The ability to save into whatever structure you want is very useful, but OK.

src/imagelist.hpp
```diff
@@ -135,21 +136,20 @@ class ImageList {
...
     /**
      * Add files from the directory to the list.
+     * @param added list for newly added entries
      * @param path path to the directory
-     * @param ordered flag to add new entry to ordered position in the list
-     * @return array with added entries
      */
-    std::vector<ImageEntryPtr> add_dir(const std::filesystem::path& path,
-                                       const bool ordered);
+    void add_dir(std::list<ImageEntryPtr>& added,
```
artemsen: What didn't work for you in the existing version? And why list instead of vector?
It is more logical to pass the return value as a return value, rather than using one of the arguments for this.
The move semantics in modern C++ makes this free.

litoj: It's all connected. I wanted to be able to return the first added image according to user sort order (see the public `add()`), same way as I was trying to change `load()`, except with `load()` I didn't realize it was working with the sources from an argument-order standpoint.

I know `std::move` etc. allows the return value to be free to save, but I don't think you can make it free to copy all the contents into another list (`add_dir()`). I did this primarily to avoid the copying for every directory level that is traversed. You can see from the benchmark that it did have some (2.8%) imact (load: 142ms->138ms). Those results, however, are with the old vector and lambda, so now it should be even faster.

It also requires less code, because you don't have to copy it to the bigger vector yourself after the `add/add_dir` calls.

The reason I've made it a `std::list` now is that its only use is to be traversed once from the beginning to the end by the caller, so a vector that would have to resize wouldn't provide any benefit.

artemsen: > avoid the copying for every directory level

Okay, that justifies the list, but let's still return the value not through the function parameters.
It is necessary to maintain a balance between optimization and code readability.

src/application.cpp
```diff
@@ -151,7 +151,7 @@ ImageEntryPtr Application::add_image_entry(const std::filesystem::path& path)
 {
     ImageList& il = ImageList::self();
 
-    const std::vector<ImageEntryPtr> entries = il.add(path);
+    std::list<ImageEntryPtr> entries = il.add(path);
```
artemsen: Can be `const`

src/imagelist.cpp
```diff
@@ -135,6 +135,17 @@ static bool compare_entries(const ImageEntry& l, const ImageEntry& r,
     return false;
 }
 
+/**
+ * Check if path is a special source.
+ * @param path path to check for special source format
+ * @return true if path starts with stdin:// or exec://
+ */
+static bool is_special_source(const std::string& path)
```
artemsen: This should belong to `ImageEntry` (`ImageEntry::is_special`).

src/imagelist.cpp
```diff
@@ -357,12 +355,12 @@ ImageEntryPtr ImageList::get(const ImageEntryPtr& from, const Dir dir)
         }
         return nullptr;
     }
-    if (dir == Dir::PrevParent) {
-        while (--it != entries.end()) {
-            if (from_parent != (*it)->path.parent_path()) {
+    if (dir == Dir::PrevParent && index > 0) {
+        do {
+            if (from_parent != (*--it)->path.parent_path()) {
```
artemsen: 1. Avoid complex expressions, `(*--it)` is unreadable.
2. If possible, use `while() {}` instead of `do{} while()`.

src/imagelist.cpp
```diff
@@ -549,33 +497,51 @@ ImageEntryPtr ImageList::add_file(const std::filesystem::path& path,
...
-        auto it = std::find_if(entries.begin(), entries.end(),
-                               [&entry, this](const ImageEntryPtr& it) {
-                                   const bool cmp =
-                                       compare_entries(*entry, *it, order);
-                                   return reverse ? !cmp : cmp;
-                               });
+
+        entry->index = it == entries.end() ? entries.size() : (*it)->index;
         entries.insert(it, entry);
+        reindex(entry->index + 1); // reindex all entries after this one
```
artemsen, line 537: Do we really have to reindex every time? Why can't we just do it once at the end of the public `add` call?

src/imagelist.hpp
```diff
@@ -176,15 +181,16 @@ class ImageList {
 
     /**
      * Reindex the image list.
+     * @param index (0-based) to start reindexing from
      */
-    void reindex();
+    void reindex(size_t index);
```
artemsen: `void ImageList::reindex(const size_t index = 0)`
(I know you need non const index, but it tells that this argument is input)

test/imagelist_test.cpp
```diff
@@ -57,19 +57,19 @@ TEST(ImageListTest, LoadFile)
...
     EXPECT_TRUE(*entry);
-    EXPECT_NE(entry->index, 0UL);
+    EXPECT_EQ(entry->index, 1UL);
     EXPECT_EQ(entry->path, "exec://2 ");
 
     entry = il.get(entry, ImageList::Dir::Next);
     ASSERT_TRUE(entry);
     EXPECT_TRUE(*entry);
-    EXPECT_NE(entry->index, 0UL);
+    EXPECT_EQ(entry->index, 2UL);
```
artemsen, line 72: I haven't looked into it yet, but it looks suspicious.
The index changed to 1 everywhere, and here to 2.

src/imagelist.hpp
litoj: This is a really unnecessary change. If you pointed out that the `sort(bool locked)` could have had a default value and be const, I'd get that because both would have made sense and introduced no reduntant code. However, here you want to add a default value, which none of the other methods declare, even though the default value would get used in max 1 place. Making it const means cluttering the code with unnecessary code and making a mess in variable names. `add_entry` also has an input parameter that is not const, because it sets its index, so it wouldn't make sense. Here it doesn't make sense either, because you cannot make a number be an output variable anyway without being a pointer.

Usually, simpler code is the correct one. It is also more readable, maintainable and therefore I'd argue - better.

test/imagelist_test.cpp
litoj: you don't need to look up beyond where the entry is coming from :D

When you call `il.get(secondentry,next)`, I'd expect you get the third entry with `index=2`.

src/imagelist.cpp
litoj: Because `bool ordered` tells you if the caller expects you to keep the entries ordered or not. And keeping the order involves also keeping their state in order, as suggests the fact that `sort()` also reindexes the items.

Another reason is clarity and consistency - given that the index is set in all the cases, it only makes sense that the other indexes would be kept in-tact.

There is no reason to be reindexing in `add` when it passes the `ordered` flag to the methods that actually affect the entries list. Purpose of `add` is to determine what type is being added, `add_entry` serves the purpose of actually adding the entry and therefore it is its responsibility to respect the `ordered` flag.

There was duplication of reindexing before, there isn't now. It is much easier to read, because all changes to the indexes are in the place that actually affect the indexes and you can immediately tell from it that it keeps the indexes consistent.

litoj (commented):
Meaningful changes have been done, others I'd prefer to discuss if they actually benefit the code overall.

src/imagelist.hpp
artemsen: > value would get used in max 1 place

Ok, let's live without default value.

> Making it const means cluttering the code

If you go through the code, you'll notice that almost all input parameters in the functions are marked as `const`, even if they are primitives and this creates additional variables in the functions. If there is a non-const input argument somewhere, then I just accidentally missed it.
This approach allows you to quickly separate parameters into input and output.
(Generally speaking, in C/C++ variables are non-const by default only for compatibility reasons; in modern languages ​​(Rust, for example), this has finally been resolved).

PS. Please, don't resolve conversations. Logically, this should be done by the person who asked the question and accepted the answer received. Otherwise, it turns out that you have unilaterally decided that the discussion is over. And it’s difficult for me to keep track of what has already been actually done and what corrections need to be reviewed.

src/image.hpp
```diff
@@ -35,6 +35,16 @@ struct ImageEntry {
     static constexpr const char* SRC_STDIN = "stdin://";
     // Special prefix used to load images from external command output
     static constexpr const char* SRC_EXEC = "exec://";
+
+    /**
+     * Check if path is a special source.
+     * @param path path to check for special source format
+     * @return true if path starts with stdin:// or exec://
+     */
+    static bool is_special_source(const std::string& path)
```
artemsen: The entire `ImageEntry` is just a wrapper around source. So, `is_special` would be enough.
And, please, move implementation to cpp.

src/imagelist.cpp
```diff
@@ -337,36 +323,40 @@ ImageEntryPtr ImageList::get(const ImageEntryPtr& from, const Dir dir)
...
             }
         }
-        return nullptr;
-    }
-    if (dir == Dir::PrevParent) {
-        while (--it != entries.end()) {
+    } else if (dir == Dir::PrevParent && index > 0) {
+        const auto start = entries.begin();
+        while (true) {
+            it--;
```
artemsen: In C++, the general rule is to prefer prefix increment (++it) over postfix increment (it++) when using iterators, unless you specifically need the old value.
It usually more efficient because no temporary object is created.

src/imagelist.cpp
```diff
@@ -337,36 +323,40 @@ ImageEntryPtr ImageList::get(const ImageEntryPtr& from, const Dir dir)
...
             }
+
+            if (it == start) {
+                break;
+            }
         }
-        return nullptr;
     }
 
-    assert(false && "unhandled iterating position");
```
artemsen, line 369: This is a fairly long function without any switch/case statements, so it's easy to miss something when refactoring it. This assert was intended to catch such cases.

src/imagelist.hpp
litoj: I tend to use it for clarifying what has been already fixed and what needs further discussion, otherwise I need to make redundant comments like this one to send a message that the requested changes are done, which I could equivalently just resolve the conversation for.

But I fully understand your POV as well. It certainly is quite annoying to have to figure out which conversation is from which point and in what state the resulting code is now.

src/imagelist.cpp
```diff
@@ -402,92 +382,57 @@ ssize_t ImageList::distance(const ImageEntryPtr& from, const ImageEntryPtr& to)
...
-            --it;
-        }
+    std::filesystem::path abs_path;
+    try {
+        abs_path = std::filesystem::absolute(path).lexically_normal();
+    } catch (...) {
+        Log::warning("Invalid path {}, skipped", path.string());
+        return {};
+    }
+    if (!std::filesystem::exists(path)) {
```
artemsen: It is better to use `abs_path` here (and further on too).

src/imagelist.cpp
```diff
@@ -402,92 +382,57 @@ ssize_t ImageList::distance(const ImageEntryPtr& from, const ImageEntryPtr& to)
...
-        path.string().starts_with(ImageEntry::SRC_EXEC)) {
-        ImageEntryPtr entry = std::make_shared<ImageEntry>();
-        entry->path = path;
-        entry->mtime = 0;
-        entry->size = 0;
-        entry->index = 0;
-        if (add_entry(entry, ordered)) {
-            entries.push_back(entry);
+    if (std::filesystem::is_directory(path)) {
+        auto ret = add_dir(abs_path);
```
artemsen: 1. Please don't overuse `auto`. They're acceptable in some cases (loops/lambdas/iterators), but don't be lazy and write the full type if it's possible. I understand that `auto` is very convenient, but such code is difficult to analyze.
2. Use meaningful names for variables. Variable `ret` with unknown type. In a couple of months you won't understand what's going on here.

src/imagelist.cpp
```diff
@@ -610,10 +569,10 @@ void ImageList::sort(bool locked)
...
 }
 
-void ImageList::reindex()
+void ImageList::reindex(const size_t index)
 {
-    size_t index = 0;
-    for (auto& it : entries) {
-        it->index = ++index;
+    // i-- to reindex also at the actual index
+    for (size_t i = entries.size(); i-- > index;) {
```
artemsen, line 576: 1. `--i`
2. You're in for a surprise if you accidentally reindex an empty array. `0 - 1 > 0` for unsigned types.

litoj: No, there is no surprise, because the whole point and existence of `x--` is so that you can compare the value before you change it. This does precisely what it is supposed to do and there are no exceptions to it =)

artemsen: Yeah, I overlooked it, sorry.

src/imagelist.cpp
```diff
@@ -402,92 +380,57 @@ ssize_t ImageList::distance(const ImageEntryPtr& from, const ImageEntryPtr& to)
...
-                entries.push_back(add_file(abs_path, ordered));
-                if (adjacent) {
-                    const std::vector<ImageEntryPtr> edir =
-                        add_dir(abs_path.parent_path(), ordered);
-                    entries.insert(entries.end(), edir.begin(), edir.end());
-                }
-            }
+        return added;
+    } else {
+        if (const ImageEntryPtr& entry = add_file(abs_path, ordered)) {
```
artemsen: 1. `add_file` doesn't return reference, but `shared_ptr`.
2. Don't skimp on lines.
```
const ImageEntryPtr entry = add_file(abs_path, ordered);
if (...
```

src/imagelist.cpp
```diff
@@ -337,28 +323,30 @@ ImageEntryPtr ImageList::get(const ImageEntryPtr& from, const Dir dir)
...
+        while (++it != end) {
             if (from_parent != (*it)->path.parent_path()) {
                 return *it;
             }
         }
         return nullptr;
-    }
-    if (dir == Dir::PrevParent) {
-        while (--it != entries.end()) {
+    } else if (dir == Dir::PrevParent) {
```
artemsen: `else` is useless, you have `return` in each already handled cases.

src/imagelist.cpp
```diff
@@ -337,28 +323,30 @@ ImageEntryPtr ImageList::get(const ImageEntryPtr& from, const Dir dir)
...
+        const auto start = entries.begin();
+        while (it-- != start) {
```
artemsen, line 350: What if the first entry has another parent and should be returned?

src/imagelist.cpp
```diff
@@ -402,92 +380,57 @@ ssize_t ImageList::distance(const ImageEntryPtr& from, const ImageEntryPtr& to)
...
-        if (*it == to) {
-            it_to = it;
-            if (it_from == entries.end()) {
-                forward = false;
-            }
+std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path,
+                                        const bool ordered)
+{
+    if (ImageEntry::is_special(path)) {
+        if (const ImageEntryPtr& entry = add_special_source(path, ordered)) {
```
artemsen: 1. `add_special_source` doesn't return reference, but `shared_ptr`.
2. Don't skimp on lines.
```
const ImageEntryPtr entry = add_file(abs_path, ordered);
if (...
```

src/imagelist.cpp
litoj: And that is precisely why I removed the `assert()` before, because it perfectly handled all the cases, whereas your requirement of keeping the `assert` means the final return cannot be used and must be duplicated multiple times in order to do the exact same thing.

Why are you trying so hard to make the code less readable by adding as many redundant lines as possible? It doesn't make anything better or clearer - or at least I don't get what benefit you see in all those changes around this chunk.

src/imagelist.cpp
litoj: I don't know if by first entry you mean `entries[0]` or `*it`. Either way, the logic hasn't changed at all.

Given that `*it` is just the input entry, there is no point in checking the iterator before iterating in the sought direction. It was done exactly the same way before - the only exception being the comparison was against the `-1`st element, because `std::list begin-1 == end`

I can add a comment to explain it in the code, if you think that's better. I certainly prefer meaningful comments over unnecessary changes of code.

src/imagelist.cpp
artemsen: > requirement of keeping the `assert` means the final return cannot be used

I don't fully understand what you mean by this.

`assert` is needed for debugging and only in case we missed something.

`else` is redundant because there is a return before it anyway (it's strange that clang-tidy skips this).

> Why are you trying so hard to make the code less readable by adding as many redundant lines as possible? It doesn't make anything better or clearer - or at least I don't get what benefit you see in all those changes around this chunk.

Are you referring to a specific case or in general?

src/imagelist.cpp
artemsen: Oh... I got it. Even for me it is a complex logic: check, then move, then check again.
Ok, let's leave it as is. I might come back here later to make it slightly clear.

src/imagelist.cpp
litoj: > I don't fully understand what you mean by this.

My original code
```cpp
if(nextparent)
  while() if() return x
else if (prevparent)
  while() if() return x

return nullptr
```

Your required redundancy caused by the assert:
```cpp
if(nextdir)
  while() if() return x
  return nullptr

if(prevdir)
  while() if() return x
  return nullptr

assert(absolutely impossible)
return nullptr
```

> Are you referring to a specific case or in general?

To this case in particular. Though, to be clear, even in this case I understand what you're saying about the method being long so you want to separate each case into a fully resolved block of code. But for that you don't need to make the code take up any more lines than are needed to achieve the same logic.

Part of the problem is the formatting that forces a simple oneline iff to be 3 lines and the iffs being on separate lines. But also, I don't see any benefit in adding more lines just to be able to write an `else if` as `return;}; if`.

PS: I splitted it as you wished. But personally I find the `else if` better because it clearly connects related blocks of code and helps to distinguish them from other unrelated checks in the code block.

artemsen: The answer is simple. You write code for the compiler. I write code for people. That's the difference =)

PR #492, opened 2026-05-25
Add support for dragging files onto the application window.

de-vri-es:
Add a new `on_file_drop` event to `AppMode`, allowing user code to handle dropped files. The list can contain 1 or more paths (but not 0, those are filtered out).

An example is included that simply opens single files dropped on the window.

Note that the bindings can't be removed. I followed the pattern of `on_image_change`, which doesn't seem to have a mechanism to remove a callback either.

It could be argued that `bind_reset` should also clear these event handlers, but as it is, that function only clears keyboard, mouse and signal bindings.

### DIFF
--- a/extra/example.lua
+++ b/extra/example.lua
@@ -159,3 +159,10 @@ swayimg.gallery.on_key("Ctrl-p", function()
     end
   end
 end)
+
+-- open a dropped file in the viewer
+swayimg.viewer.on_file_drop(function(paths)
+  if #paths == 1 then
+    swayimg.viewer.open(paths[1])
+  end
+end)
--- a/extra/swayimg.lua
+++ b/extra/swayimg.lua
@@ -453,6 +453,11 @@ function swayimg_appmode.on_signal(signal, fn) end
 ---@param fn function Handler for notifications about changing the current image
 function swayimg_appmode.on_image_change(fn) end
 
+---Add a callback function called when files are dropped on the window.
+---Since 5.2.
+---@param fn fun(paths: string[]) Handler receiving an array of dropped file paths
+function swayimg_appmode.on_file_drop(fn) end
+
 ---Set text layer scheme.
 ---Since 5.0.
 ---@param pos block_position_t Text block position
--- a/src/appevent.hpp
+++ b/src/appevent.hpp
@@ -9,7 +9,9 @@
 
 #include <filesystem>
 #include <functional>
+#include <string>
 #include <variant>
+#include <vector>
 
 /** Application event. */
 namespace AppEvent {
@@ -73,6 +75,11 @@ struct FileRemove {
     std::filesystem::path path; ///< Path to the file
 };
 
+/** File drop event (drag-and-drop receive). */
+struct FileDrop {
+    std::vector<std::string> paths; ///< Dropped file paths
+};
+
 // clang-format off
 using Holder = std::variant<WindowClose,
                             WindowRedraw,
@@ -85,7 +92,8 @@ using Holder = std::variant<WindowClose,
                             Signal,
                             FileCreate,
                             FileModify,
-                            FileRemove>;
+                            FileRemove,
+                            FileDrop>;
 // clang-format on
 
 // event handler
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -403,6 +403,9 @@ void Application::handle_event(const AppEvent::Holder& event)
             } else if constexpr (std::is_same_v<decltype(event),
                                                 const AppEvent::FileRemove&>) {
                 handle_event(event);
+            } else if constexpr (std::is_same_v<decltype(event),
+                                                const AppEvent::FileDrop&>) {
+                handle_event(event);
             } else {
                 assert(false && "unhnadled event type");
                 handle_event(event);
@@ -511,6 +514,11 @@ void Application::handle_event(const AppEvent::FileRemove& event)
     if (!std::filesystem::is_directory(event.path)) {
         remove_image_entry(event.path);
     }
+}
+
+void Application::handle_event(const AppEvent::FileDrop& event)
+{
+    current_mode()->notify_file_drop(event.paths);
 }
 
 void Application::signal_handler(int signal)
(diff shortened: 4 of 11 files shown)


artemsen:
Nice!
But this is an image viewer, not a file manager, so there's no point in processing the list in Lua. It would be more logical to simply add files to the image list and open the first one submitted.
This will make it possible to use this feature out of the box without changing the configs.

Let's remove the Lua API extension?

rempas:
> Nice! But this is an image viewer, not a file manager, so there's no point in processing the list in Lua. It would be more logical to simply add files to the image list and open the first one submitted. This will make it possible to use this feature out of the box without changing the configs.
> 
> Let's remove the Lua API extension?

What if the user wants to do something with the dropped files tho? Maybe not open files that are already opened? Maybe only want to open up to a certain number of files? I don't think that it's bad to have a way to let the users choose, while having the default option to just add the files in the end of the list (which I think is what you also propose, tho why you said just the first one submitted?).

What you don't like is that you want the feature to be usable by default? Or is there something else that you don't like? Because if it's the first case, it's possible to have a default behavior, right?

de-vri-es:
> Nice! But this is an image viewer, not a file manager, so there's no point in processing the list in Lua. It would be more logical to simply add files to the image list and open the first one submitted. This will make it possible to use this feature out of the box without changing the configs.
> 
> Let's remove the Lua API extension?

I think there's some merit for letting the user write the handler. For example, I have `imagelist.enable_adjecent(true)`, and I want to clear any old images in the list before opening the dropped image. That way, it's essentially as if the application has just been run on the new image, deleting any old state. (I think I want to add an `imagelist.clear()` function too :innocent: )

This could also be tweaked by boolean config option, but I figured the Lua handler is more flexible and actually requires less Lua exposed items (like `swayimage.viewer.clear_image_list_on_file_drop(true)` or maybe `swayimage.imagelist.clear_on_file_drop(true)`?).


/edit:

To make it more clear, my handler would look like this, if `imagelist.clear()` existed:


```lua
function dirname(str)
    return str:match("(.*/)")
end

swayimg.viewer.on_file_drop(function(paths)
	swayimg.imagelist.clear()
	swayimg.imagelist.add(dirname(paths[1]))
	swayimg.viewer.open(paths[1])
end)
```

artemsen:
I'm not against the Lua handler, but...
1. We need a default handler. As I see, the simplest option is to add everything to the image list and show the first one.
2. The current design of the application dies not allow for a simple solution with `imagelist.clear()` (it will end up with "No more images, exit"). swayimg_appmode.on_file_drop doesn't look very useful without it, especially if we have default handler.
3. Lua API requires separate discussion. I'd ask you to split the PR into two: drag-and-drop support and Lua API extension. I'm ready to merge the first one into master.

litoj:
> The current design of the application dies not allow for a simple solution with imagelist.clear() (it will end up with "No more images, exit"). swayimg_appmode.on_file_drop doesn't look very useful without it, especially if we have default handler.

I know it will still take some time before the other PRs leading to my `imagelist.set()` implementation will get merged, but this is precisely the case where it would come in handy - it's exactly the _clear and add_ behaviour, but it avoids having to deal with the imagelist being empty for a moment.

In other words, I already have [a solution](https://github.com/litoj/swayimg/blob/bulk-list-update/src/imagelist.cpp#L242) that also provides a signifficant performance boost without having to deal with an empty imagelist.

de-vri-es:
> I'm not against the Lua handler, but...
> 
> 1. We need a default handler. As I see, the simplest option is to add everything to the image list and show the first one.
> 2. The current design of the application dies not allow for a simple solution with `imagelist.clear()` (it will end up with "No more images, exit"). swayimg_appmode.on_file_drop doesn't look very useful without it, especially if we have default handler.
> 3. Lua API requires separate discussion. I'd ask you to split the PR into two: drag-and-drop support and Lua API extension. I'm ready to merge the first one into master.

Fair. We can start with a default handler.

I did find one problem while testing: if you drag-and-drop to the same window, the application hangs. I think we need to ignore drops to/from the same surface, but I haven't yet found a way to check the source surface in the drop receiver, or the target surface in the drop provider. It may just need to be a local flag to remember we started a drag-and-drop, and ignore all drops while that flag is set :(

Regarding 2, I did notice that while implementing `clear`. So I was instead considering something like `imagelist.set_entries(std::vector<std::string> new entries)`. I guess that's essentially the same solution as @litoj has in mind.

de-vri-es:
Although it be more efficient to instead allow the imagelist to be empty. Then instead of firing `removed` events per removed image, you can just throw away all cached state and rebuild it for the new image list once something is added.

artemsen:
> application hangs

This is a blocking problem.

> allow the imagelist to be empty

What should the viewer show at this point? A blank window?

de-vri-es:
> > application hangs
> 
> This is a blocking problem.

Yeah, I want to dig into the wayland protocol a bit more before resorting to a local flag that we need to keep in sync across events. I'll update the PR with a fix once I've either found it or given up and resorted to a local flag.

> > allow the imagelist to be empty
> 
> What should the viewer show at this point? A blank window?

Yeah, I imagine so.

artemsen:
> application hangs

Apparently, the problem was not solved?
The app still hangs when trying to drag an image onto itself.

de-vri-es:
Yeah, didn't have time to continue until now. Just rebased before fixing it.

de-vri-es:
I didn't find a way to detect drag-and-drop to self other than keeping a flag locally to remember we're already initiated a drag-and-drop.

So that's the solution now.

Still need to add a default handler.

de-vri-es:
Hmm, something broke when pulling in latest master. Need to figure out what.

/edit: Fixed.

artemsen:
I've slightly refactored your code and merged it into the master. By default, files are added to an existing list, and there's no lua support yet, as it's irrelevant without imagelist.clear . But I'm working on it.
Thank you!

de-vri-es:
Thanks for putting in the extra effort!

PR #494, opened 2026-05-27
refactor: mark as const where appropriate

litoj:
I can't do anything half-way, so I made sure that the policy of marking input parameters as `const` holds everywhere, where possible and appropriate.

### DIFF
--- a/src/formats/avif.cpp
+++ b/src/formats/avif.cpp
@@ -19,7 +19,7 @@ public:
     using AvifDecoder =
         std::unique_ptr<avifDecoder, decltype(&avifDecoderDestroy)>;
 
-    ImagePtr decode(const Data& data) override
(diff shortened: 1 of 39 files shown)


src/formats/avif.cpp
```diff
@@ -19,7 +19,7 @@ class ImageFormatAvif : public ImageFormat {
     using AvifDecoder =
         std::unique_ptr<avifDecoder, decltype(&avifDecoderDestroy)>;
 
-    ImagePtr decode(const Data& data) override
+    [[nodiscard]] ImagePtr decode(const Data& data) const override
```
artemsen, line 22: This was intentionally left non-constant, since some libraries require initialization and this can only be done once, and not for each file opening. But I never got around to it =)

PR #496, opened 2026-05-28
Allow bulk imagelist updates

litoj:
Implementation of bulk list changes that allows swapping the entire imagelist for a set of different sources.

Improves supportability of #481 

The current implementation is the simplest way of supporting directories as well as files. Supporting just one wouldn't change much - other than removing the dir iterator.

I also fixed the FsMonitor to stop doing those linear lookups that were having signifficant impact on startup performance because of the need to check every single file when adding it. Now the startup takes about 1/3 of the time =)

PS: sorry for the large change. I tried multiple approaches to implementing this function and I kept running into duplicating a lot of code, so I ended up progressively adding extra changes to the underlying structure to allow a cleaner implementation (still not the prettiest, but there aren't that many options to make it better).

I have [an alternative implementation](https://github.com/litoj/swayimg/blob/fa9d42f357a7682316fef05696030be5652f84e0/src/imagelist.cpp#L231-L314), that avoids the inlined lambdas and adds the new files at the end instead, but it is slower.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -57,6 +57,7 @@ or in the file `/usr/share/swayimg/example.lua` after installing the program.
 * Image list
   * [swayimg.imagelist.size](#swayimgimagelistsize): Get number of entries in the image list
   * [swayimg.imagelist.get](#swayimgimagelistget): Get list of all entries in the image list
+  * [swayimg.imagelist.set](#swayimgimagelistset): Set list of images to given entries
   * [swayimg.imagelist.add](#swayimgimagelistadd): Add entry to the image list
   * [swayimg.imagelist.remove](#swayimgimagelistremove): Remove entry from the image list
   * [swayimg.imagelist.set_order](#swayimgimagelistset_order): Set sort order of the image list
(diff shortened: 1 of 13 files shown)


artemsen:
> PS: sorry for the large change. I tried multiple approaches to implementing this function and I kept running into duplicating a lot of code, so I ended up progressively adding extra changes to the underlying structure to allow a cleaner implementation (still not the prettiest, but there aren't that many options to make it better).

Please move all code not related to "Allow bulk imagelist updates" to a separate PR. It's better to do a "dumb" implementation and gradually improve it than to do everything at once and get caught up in discussions that are not related to the topic of this PR.

litoj:
Now there should really be only the necessary changes. (all of #500 included)

artemsen:
> (all of #500 included)

I've already expressed my opinion on these changes. Don't try to bruteforce me, it won't change my mind.

But overall I don't like the solution with the tracker. I think we need to fix the problem, not try to treat the symptoms. What I mean is that `imagelist.clear` needs to be implemented and I've started moving in that direction.

litoj:
Allright. I still don't have your opinion on the other slower approach, though. Let's give that one a try.

Il.clear() will help for sure, but it still has the flaw of adding images one-by-one which will be quadratic when order!=none and the constant locking and unlocking and calling of the mode handlers will still be a notable overhead.

More importantly, by clearing everything, you also discard all already loaded thumbnails, so even if the images haven't all changed, this will reload them all.

litoj:
Since it is simply impossible to make this behaviour work also for directories without duplicating the entire recursive directory iteration logic and also without using a lambda for its iteration, I am closing this.

But it should be known that the primary difference between clear and add vs set is that the cached images won't have to be reloaded with my aproach and the switch will be much faster because all images will be added at once. So if you have any suggestions for how I should go about doing it that would work for you as well, I'd be happy to give it another try =)

PR #497, opened 2026-05-29
fsmonitor: Improve lookup performance

litoj:
Cuts image loading time in half:
`@master: Image list loaded in 0.11`
`@this: Image list loaded in 0.6`

### DIFF
--- a/src/fsmonitor.cpp
+++ b/src/fsmonitor.cpp
@@ -81,15 +81,10 @@ void FsMonitor::add(const std::filesystem::path& path)
     assert(path.is_absolute());
 
     // check if file is already watched by its parent
-    if (std::filesystem::is_regular_file(path)) {
-        const std::filesystem::path& parent = path.parent_path();
-        const auto it =
-            std::find_if(watch.begin(), watch.end(), [&parent](const auto& it) {
-                return parent == it.second.parent_path();
-            });
-        if (it != watch.end()) {
-            return; // already watched
-        }
+    if (watched.contains(std::filesystem::is_regular_file(path)
+                             ? path.parent_path()
+                             : path)) {
+        return;
     }
 
     const int wd =
@@ -102,21 +97,20 @@ void FsMonitor::add(const std::filesystem::path& path)
     }
 
     watch.insert(std::make_pair(wd, path));
+    watched.insert(path);
 }
 
 void FsMonitor::handle_event(const inotify_event* event)
 {
+    const auto it = watch.find(event->wd);
     if (event->mask & IN_IGNORED) {
         // remove from the watch list
-        watch.erase(event->wd);
+        watched.erase(it->second);
+        watch.erase(it);
         return;
     }
 
-    const auto it = watch.find(event->wd);
     assert(it != watch.end());
-    if (it == watch.end()) {
-        return;
-    }
 
     // compose full path
     std::filesystem::path path = it->second;
--- a/src/fsmonitor.hpp
+++ b/src/fsmonitor.hpp
@@ -5,7 +5,8 @@
 #pragma once
 
 #include <filesystem>
-#include <map>
+#include <unordered_map>
+#include <unordered_set>
 
 struct inotify_event;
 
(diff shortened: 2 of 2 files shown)


artemsen:
> Cuts image loading time in half:
> `@master: Image list loaded in 0.11`
> `@this: Image list loaded in 0.6`

So, you've made things slower. Profit! =)

litoj:
> So, you've made things slower. Profit! =)

Lol, thanks xD. But I think you know what I meant.

This time I'm not gonna write it by hand but just copy the --verbose lines:

- Yours: Image list loaded in 0.189459 sec
- Mine: Image list loaded in 0.112694 sec

0.19 and 0.11 were both very consistent results for a deeper structured directory with 2300 files. For a flat directory the results were indistinguishable.

artemsen:
Typically, these tests take into account not only speed, but also memory consumption. Lost kilobytes probably aren't that important, but still.

There are two things I don't like the most:
1. Duplicate information - the same path is now stored in two different containers.
2. The containers are incoherent. It's hypothetical that one could be changed but the other forgotten.

So saving 0.08 seconds on startup doesn't seem like a compelling reason to make such a change.

litoj:
> So saving 0.08 seconds on startup doesn't seem like a compelling reason to make such a change.

This is not just about startup time, this is about any additions to the image list. This literally halves the time it takes.

Yes, there is extra memory consumed. On the other hand, given how rare file changes are, we could just flip the map around and do a linear search for the file descriptor when it changes instead. That way you wouldn't need a second structure.

I'd also like to point out, that this is the same situation as in imagelist - there are also 2 structures where one is for order and one for lookups. Same with gallery - `cache`, `active`... So I really wouldn't focus on forgetting something. We are humans, it will happen whether we try to avoid it or not (just like that `i>0` that you missed in the `get()` refactor). Compromising very signifficant speed improvement for a very small change in an otherwise relatively simple file seems unreasonable.

artemsen:
> I'd also like to point out, that this is the same situation as in imagelist

Yeah, I don't like that either, but at least it gives a real speed boost.

> Compromising very signifficant speed improvement for a very small change in an otherwise relatively simple file seems unreasonable.

0.08 sec is not "very signifficant speed improvement".

litoj:
> 0.08 sec is not "very signifficant speed improvement"

It is not about the total time, it is about the percentage and the situations in which it will be used.

Again - when set_entries gets added - changing a directory or filtering will mean all the files will get re-added and therefore the same stutter as happens on initialization will happen in the UI, except now the user will feel it.

litoj:
When discussing the Gallery changes, you prefered a 0.1ms improvement in redraw over a 1ms improvement in thumbnail loading. Here, the speed difference is >100x that.

Just to note - the memory difference is almost 0 because when you load a directory, this will load the dir and all subdirs, but not the files, meaning comparatively it will still be maybe 20 folders vs 2000 files that won't be in there.

artemsen:
> Just to note - the memory difference is almost 0 because when you load a directory, this will load the dir and all subdirs, but not the files, meaning comparatively it will still be maybe 20 folders vs 2000 files that won't be in there.

So, it's not even 0.08 sec, but several times less.

litoj:
> So, it's not even 0.08 sec, but several times less.

No, XD? How did you manage to come to that conclusion? The numbers are real results.

In every add() every added file (and dir, but those are just few) has to get registered. Currently, it checks for registration by going through all already tracked entries. So through every directory and every file that was added manually or passed in on startup. That means it is bogously slow for both large structured directories as well as viewing a selection of files.

Also, the memory diff would be cca 1kB for every 19 folders or 16 selected files (assuming 50 chars per dir path).

So yes, memory isn't really an issue and performance isn't either, unless you enjoy snappy application startup (I know I do) or intend to switch between directories.

It's a small overall benefit for a small overall change that also simplifies the ckecking proces. So I am still struggling to underatand why it is such a problem );

It's not like there is a lot going in in that file anyway, so chances for mistakes are lower as well (compared to imagelist etc.)

PR #498, opened 2026-05-29
Minor fixes

litoj:
Deduplicated some code in preparation for adding `set_entries` and removed unreachable cases in `viewer:switch_image`

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -159,15 +159,7 @@ ImageEntryPtr Application::add_image_entry(const std::filesystem::path& path)
         current_mode()->handle_imagelist(AppMode::ImageListEvent::Create, it);
     }
 
-    // update image list text info
-    Text& text = Text::self();
-    text.set_field(Text::FIELD_LIST_INDEX,
-                   std::to_string(current_mode()->current_entry()->index + 1));
-    text.set_field(Text::FIELD_LIST_TOTAL, std::to_string(il.size()));
-    text.update();
-
-    redraw();
-
+    Text::self().update_list_info();
     return entries.front();
 }
 
@@ -181,15 +173,7 @@ void Application::remove_image_entry(const std::filesystem::path& path)
 
     il.remove(entry);
     current_mode()->handle_imagelist(AppMode::ImageListEvent::Remove, entry);
-
-    // update image list text info
-    Text& text = Text::self();
-    text.set_field(Text::FIELD_LIST_INDEX,
-                   std::to_string(current_mode()->current_entry()->index + 1));
-    text.set_field(Text::FIELD_LIST_TOTAL, std::to_string(il.size()));
-    text.update();
-
-    redraw();
+    Text::self().update_list_info();
 }
 
 void Application::add_fdpoll(const int fd, const FdEventHandler& handler)
--- a/src/gallery.cpp
+++ b/src/gallery.cpp
@@ -556,7 +556,7 @@ void Gallery::clear_invisible_thumbnails()
         const size_t visible_last = scheme.back().img->index;
         const size_t store_min =
             visible_first > cache_size / 2 ? visible_first - cache_size / 2 : 1;
-        const size_t store_max = visible_last + cache_size - cache_size / 2;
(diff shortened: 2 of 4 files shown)


src/viewer.cpp
```diff
@@ -629,19 +629,9 @@ bool Viewer::switch_image(const ImageList::Dir dir, const ImageEntryPtr& entry)
 
     while (true) {
         if (!next_entry && imagelist_loop) {
-            // reshuffle random on new loop
```
artemsen: There is even a commentary here about why this is being done.

> Random order means `il:get()` will always return a random result

No, it's not. Next returns _next_ entry, event if list is randomized.

litoj: Huh, I see now. I was seeing `Order::Random`, but actually reading `Dir::Random`.

src/application.cpp
```diff
@@ -147,6 +147,18 @@ AppMode* Application::current_mode()
     return nullptr;
 }
 
+void Application::update_list_text_info()
```
artemsen: I think this should belong to the `Text` class.

src/gallery.cpp
```diff
@@ -556,7 +556,7 @@ void Gallery::clear_invisible_thumbnails()
         const size_t visible_last = scheme.back().img->index;
         const size_t store_min =
             visible_first > cache_size / 2 ? visible_first - cache_size / 2 : 1;
-        const size_t store_max = visible_last + cache_size - cache_size / 2;
+        const size_t store_max = visible_last + (cache_size + 1) / 2;
```
artemsen, line 559: I don't understand how it relates to "text updates", but ok.

litoj: It doesn't, I just noticed it a long time ago and I was always shifting it between commits and didn't want to make a one-line commit.

src/text.cpp
```diff
@@ -229,6 +229,17 @@ void Text::reset(const ImageEntryPtr& entry)
     update();
 }
 
+void Text::update_list_info()
+{
+    Application& app = Application::self();
+    const ImageEntryPtr& entry = app.current_mode()->current_entry();
+    set_field(Text::FIELD_LIST_INDEX, std::to_string(entry->index + 1));
```
artemsen, line 236: Now we have duplicated code with in `Text::reset(const ImageEntryPtr& entry)`.

src/text.cpp
```diff
@@ -229,6 +229,17 @@ void Text::reset(const ImageEntryPtr& entry)
...
+    set_field(Text::FIELD_LIST_TOTAL, std::to_string(ImageList::self().size()));
+    update();
+
+    app.redraw();
```
artemsen, line 240: Text layer shouldn't redraw by itself (except status change).

src/text.cpp
```diff
@@ -229,6 +229,17 @@ void Text::reset(const ImageEntryPtr& entry)
     update();
 }
 
+void Text::update_list_info()
+{
+    Application& app = Application::self();
+    const ImageEntryPtr& entry = app.current_mode()->current_entry();
```
artemsen, line 235: I don't like this. Now Text goes to modes, which is not good and looks like a cyclic dependency.
It seems to me like you're tilting at windmills.
This minor copypaste wasn't as much of a problem as it became after it was corrected.

litoj: I agree, I also didn't want it in the text layer particularly for this reason.

But given that when `set_entries` is added, there will be this same code copied 3 times, I think it should be a separate method. It was easy enough to overlook the index there that I had to change to +1 when refactoring to 0-based.

src/text.cpp
litoj: Except there cannot be any new duplication since the code hasn't changed, only its placement has. So if there were duplicates before, then there are obviously still now.

Yes, it would be better to keep it in Application instead of Text. I would be happy to move it back. But I am guessing you want the same code repeated 3 times over a separate method dedicated for it, if so, then just close this PR so I don't have to keep trying things you know before-hand won't be acceptably by your standards anyway :|

PR #499, opened 2026-05-29
Consistent locking style

litoj:
Prefer using scoped_lock as it is in most of the codebase. Also prefer using scoped_lock and arbitrary code block over mutex.lock/unlock.

There is a deadlock that I am trying to get rid of. Ensuring that gallery is consistent in _not_ locking calls to `Application::redraw()` unfortunately didn't resolve it.

So I did what I could to make everything as clear as possible in the process in case I'd discover more stuff like that locked Application::redraw call.

### DIFF
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -147,13 +147,13 @@ ImageEntryPtr ImageList::load(const std::vector<std::filesystem::path>& sources)
         return nullptr;
     }
 
-    mutex.lock();
-    entries_arr.reserve(sources.size());
-    for (auto& it : sources) {
-        add(it, false);
+    {
+        const std::scoped_lock lock(mutex);
+        for (auto& it : sources) {
+            add(it, false);
+        }
+        sort();
     }
-    sort();
-    mutex.unlock();
 
     // disable loading of adjacent files, otherwise fs mon will add unnecessary
     // files to the list
@@ -211,7 +211,7 @@ ImageEntryPtr ImageList::load(const std::filesystem::path& list_file)
 
 std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path)
 {
-    const std::unique_lock lock(mutex);
+    const std::scoped_lock lock(mutex);
     return add(path, true);
 }
 
@@ -223,7 +223,7 @@ ImageEntryPtr ImageList::remove(const ImageEntryPtr& entry, const bool forward)
 
     ImageEntryPtr next = get(entry, forward ? Dir::Next : Dir::Prev);
 
-    const std::unique_lock lock(mutex);
+    const std::scoped_lock lock(mutex);
 
     entries_map.erase(entry->path);
     entries_arr.erase(entries_arr.begin() + entry->index);
@@ -244,7 +244,7 @@ void ImageList::set_order(const Order new_order)
 {
     if (order != new_order || new_order == Order::Random) {
         order = new_order;
-        const std::unique_lock lock(mutex);
+        const std::scoped_lock lock(mutex);
         sort();
     }
 }
@@ -253,7 +253,7 @@ void ImageList::set_reverse(const bool enable)
 {
     if (reverse != enable) {
         reverse = enable;
-        const std::unique_lock lock(mutex);
+        const std::scoped_lock lock(mutex);
         sort();
     }
 }
--- a/src/threadpool.hpp
+++ b/src/threadpool.hpp
@@ -51,10 +51,11 @@ public:
         auto task_fn =
             std::bind(std::forward<F>(fn), std::forward<Args>(args)...);
 
-        mutex.lock();
-        task_id = ++last_id;
-        tasks.emplace_back(Task { task_id, task_fn });
-        mutex.unlock();
+        {
+            const std::scoped_lock lock(mutex);
+            task_id = ++last_id;
+            tasks.emplace_back(Task { task_id, task_fn });
+        }
 
         tnotify.notify_one();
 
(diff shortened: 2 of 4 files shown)


src/gallery.cpp
```diff
@@ -209,9 +209,10 @@ void Gallery::reload()
     tpool.cancel();
     tpool.wait();
 
-    // clear cache
-    const std::scoped_lock lock(mutex);
-    cache.clear();
+    {
```
artemsen: For what?

src/gallery.cpp
```diff
@@ -380,7 +381,6 @@ void Gallery::handle_imagelist(const ImageListEvent event,
                                const ImageEntryPtr& entry)
 {
     if (event == ImageListEvent::Modify || event == ImageListEvent::Remove) {
-        // remove entry from cache
```
artemsen, line 385: I understand it's not really needed here. But why did you delete it? Does it break locking style?

src/gallery.cpp
```diff
@@ -599,12 +599,13 @@ void Gallery::load_thumbnail(const ImageEntryPtr& entry)
         }
     }
 
-    const std::scoped_lock lock(mutex);
```
artemsen, line 604: For what?

src/imagelist.cpp
```diff
@@ -205,7 +206,7 @@ ImageEntryPtr ImageList::load(const std::filesystem::path& list_file)
 
 std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path)
 {
-    const std::unique_lock lock(mutex);
```
artemsen, line 214: I don't consider myself a big expert in modern C++, so I rely more on the documentation, which clearly states that shared_mutex uses unique_lock for writing.

artemsen:
> There is a deadlock that I am trying to get rid of.

If this is not a result of this PR, please create a bug report.

src/imagelist.cpp
litoj: Sure, it provides an example that uses unique lock. That, however, doesn't mean that scoped lock cannot be used.

Given that both just call `.lock()` and `.unlock()` on the mutex, it doesn't matter what type the mutex is as long as it is an instance of Lockable.

src/gallery.cpp
litoj: To _not_ lock the `Application::redraw()` call, because it is not being locked anywhere else and could lead to deadlock if there was a similar mistake that the call of this function was still locked by the event loop.

Yes, it is not the case, on the other hand this entire PR is purely about improving consistency, so making this not lock the thing that doesn't get locked anywhere else seemed the right call here

src/gallery.cpp
litoj: Because redundant comments only make the code longer, which decreases the amount of context one can see.

src/gallery.cpp
litoj: To not lock the `Application::redraw()` call.

src/gallery.cpp
artemsen: > To _not_ lock the `Application::redraw()` call, because it is not being locked anywhere else and could lead to deadlock if there was a similar mistake that the call of this function was still locked by the event loop.

Look again. `Application::redraw()` uses own mutex, so this change is useless.

src/gallery.cpp
artemsen: How does this relate to "Consistent locking style"?

src/gallery.cpp
```diff
@@ -599,12 +599,13 @@ void Gallery::load_thumbnail(const ImageEntryPtr& entry)
...
-
-    if (pm) {
-        cache.insert_or_assign(entry, pm);
+    {
```
artemsen: In [application.cpp](https://github.com/artemsen/swayimg/pull/499/changes/BASE..c8e96c4a2e9a9e55631f8726cf0bfc3f0b5a32da#diff-1ac21b1b653c1f5652e24f3c5dcf82642d12ebd4dcb45e55038e827c95808d3a), they did the exact opposite. It's time to establish a consistent style.

src/gallery.cpp
litoj: I already said that it is not necessary. I am aware, but as I said - it is better to be consistent about locking than to end up with the gallery call locking the add of an event all while the call to gallery handler is already locked by the app event handler. Again - in theory.

This is just trying to avoid possible confusion and mistakes in the future.

src/gallery.cpp
litoj: that everything uses the same type of lock unless a different type is actually needed. Also, `scoped_lock` is a simpler object than `unique_lock` and the name `scoped` makes it more obvious that it locks for the duration of the code block.

PS: nevermind, I saw the comment under a different line of code. I will put it back.

src/gallery.cpp
litoj: Yup, agreed. As you can see from the code snippet in the created issue, I already fixed it in there but apparently didn't commit it.

I will change it now with the rebase.

src/threadpool.hpp
```diff
@@ -44,17 +44,18 @@ class ThreadPool {
      * @return task id
      */
     template <typename F, typename... Args> size_t add(F&& fn, Args&&... args)
-    {
+    { // C++ is dumb and doesn't resolve this if inside threadpool.cpp
```
artemsen: What did you mean by this comment?

litoj: That the method implementation cannot be put in `threadpool.cpp` because it wouldn't compile for some reason.

I should probably change the comment to that, rather than what I have there now, right? :D

artemsen: > it wouldn't compile for some reason.

This is a template class-member function, you can not move it to cpp.

PR #500, opened 2026-05-29
Separate dir iteration and skip redundant fsmonitor calls

litoj:
Generalizes recursive directory iteration to be reusable by more than just one method.

### DIFF
--- a/src/image.cpp
+++ b/src/image.cpp
@@ -28,6 +28,38 @@ bool ImageEntry::is_special(const std::string& path)
         path.starts_with(ImageEntry::SRC_EXEC);
 }
 
+ImageEntryPtr ImageEntry::from_special(const std::filesystem::path& path)
+{
+    assert(ImageEntry::is_special(path));
+
+    ImageEntryPtr entry = std::make_shared<ImageEntry>();
+    entry->path = path;
+    entry->mtime = 0;
+    entry->size = 0;
+    entry->index = 0;
+    return entry;
+}
+
+ImageEntryPtr ImageEntry::from_file(const std::filesystem::path& path)
+{
+    assert(path.is_absolute());
+    assert(std::filesystem::is_regular_file(path));
+
+    const auto fs_time = std::filesystem::last_write_time(path);
+    auto sys_time =
+        std::chrono::time_point_cast<std::chrono::system_clock::duration>(
+            fs_time - std::filesystem::file_time_type::clock::now() +
+            std::chrono::system_clock::now());
+    const std::time_t tt_time = std::chrono::system_clock::to_time_t(sys_time);
+
+    ImageEntryPtr entry = std::make_shared<ImageEntry>();
+    entry->path = path;
+    entry->mtime = tt_time;
+    entry->size = std::filesystem::file_size(path);
+    entry->index = 0;
+    return entry;
+}
+
 void Image::draw(const size_t frame, Pixmap& target, const double scale,
                  const ssize_t x, const ssize_t y)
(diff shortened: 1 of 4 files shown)


src/image.cpp
```diff
@@ -28,6 +28,38 @@ bool ImageEntry::is_special(const std::string& path)
         path.starts_with(ImageEntry::SRC_EXEC);
 }
 
+ImageEntryPtr ImageEntry::from_special(const std::filesystem::path& path)
```
artemsen, line 31: If you had done this in the constructor, I could justify moving the code out by wanting to make it "beautiful in the C++ style."
But you simply took the code away and thereby spread the logic of working with files across different places.
Do you really think this is an improvement?

src/imagelist.cpp
```diff
@@ -407,102 +429,35 @@ std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path,
...
-        if (!added.empty() && ordered) {
-            sort(false);
-        }
-        return added;
-    } else if (adjacent) {
-        std::list<ImageEntryPtr> added = add_dir(abs_path.parent_path());
+    std::list<ImageEntryPtr> added;
+    const bool is_dir = std::filesystem::is_directory(abs_path);
+    if (is_dir || adjacent) {
+        iterate_dir(
```
artemsen, line 442: I don't share your love for callbacks.

src/image.cpp
litoj: I haven't even thought of that. That would be better for sure.

It would have only one downside - having to check for special path for every added entry, which would make each adding of an entry probably half as fast.

My main intention is to separate adding an entry and creating one, so that I can work with the entries more efficiently in the bulk_update (prepare the new ones first, then just add all existing and new entries in the source order).

An extra goal was taking `fsmon.add(file)` out of the `add_file` so that it doesn't try to add every file in the directory when iterating over a directory, but rather adds the file to fsmon only when the file path was directly passed to `add`. That way all unnecessary calls to fsmon are avoided.

I have an alternate set_entries implementation that does everything with the paths first and then adds them, in which case this could be ignored - but it is slower and also redundantly checks the entry map when we know all entries added are new, so... I'd rather try to find an acceptable version of this ```)

src/imagelist.cpp
litoj: I know. I tried creating an iterator that would go through the inner directories as well, but the boilerplate for it is like 120 lines, so it really didn't fit. This is a simple solution and lends itself to checking the directories without having to add the found entries.

I presume you'd prefer a solution in the style of `list<path> collect_paths(source, recursive, fsmon)`, right?

That would work as well. But it would require collecting all the paths first and then looping over them again, which would duplicate a lot of unnecessary data and because of creating every object separately it would also be slower because of all the small individual memory allocations.

Realistically, this is the most performat _and_ simplest and cleanset way it can be done in, other than duplicating the entire code for the recursive iteration.

src/image.cpp
artemsen: > My main intention is to separate adding an entry and creating one, so that I can work with the entries more efficiently in the bulk_update (prepare the new ones first, then just add all existing and new entries in the source order).

Don't write code "for later." That way, you'll never finish it.
It may turn out that it shouldn't be like that at all.

src/imagelist.cpp
artemsen: > This is a simple solution

This is not a _simple_ solution. These things are easy to write, but difficult to maintain.

PR #505, opened 2026-06-03
imagelist: Unify load and add logic

litoj:
Improves performance of list adding and removing for multiple files and streamlines the methods to have the same arguments and return values.

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -231,16 +231,16 @@ ImageEntryPtr Application::il_initialize()
     }
     if (!sparams.sources.empty()) {
         if (sparams.sources.size() == 1 && sparams.sources[0] == "-") {
-            first_entry = il.load(std::vector<std::filesystem::path> {
-                                      ImageEntry::SRC_STDIN })
+            first_entry = il.add(std::vector<std::filesystem::path> {
+                                     ImageEntry::SRC_STDIN })
                               .front();
         } else {
-            first_entry = il.load(sparams.sources).front();
+            first_entry = il.add(sparams.sources).front();
         }
     }
     if (sparams.from_file.empty() && sparams.sources.empty()) {
         first_entry =
-            il.load(std::vector<std::filesystem::path> { "." }).front();
+            il.add(std::vector<std::filesystem::path> { "." }).front();
     }
 
     if (Log::verbose_enable()) {
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -231,14 +231,16 @@ ImageEntryPtr Application::il_initialize()
     }
     if (!sparams.sources.empty()) {
         if (sparams.sources.size() == 1 && sparams.sources[0] == "-") {
-            first_entry = il.load(
-                std::vector<std::filesystem::path> { ImageEntry::SRC_STDIN });
+            first_entry = il.load(std::vector<std::filesystem::path> {
+                                      ImageEntry::SRC_STDIN })
+                              .front();
         } else {
-            first_entry = il.load(sparams.sources);
+            first_entry = il.load(sparams.sources).front();
         }
     }
     if (sparams.from_file.empty() && sparams.sources.empty()) {
-        first_entry = il.load(std::vector<std::filesystem::path> { "." });
+        first_entry =
+            il.load(std::vector<std::filesystem::path> { "." }).front();
     }
 
     if (Log::verbose_enable()) {
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -152,10 +152,7 @@ Application::add_images(const std::vector<std::filesystem::path>& paths)
 {
     ImageList& il = ImageList::self();
 
-    std::list<ImageEntryPtr> entries;
-    for (const auto& path : paths) {
-        entries.splice(entries.end(), il.add(path));
-    }
+    const std::list<ImageEntryPtr> entries = il.add(paths);
 
     if (!entries.empty()) {
         current_mode()->handle_imagelist(AppMode::ImageListEvent::Create,
@@ -246,6 +243,10 @@ ImageEntryPtr Application::il_initialize()
     if (Log::verbose_enable()) {
         Log::verbose("Image list loaded in {:.6f} sec", timer.time());
     }
+
+    // disable loading of adjacent files, otherwise fs mon will add unnecessary
+    // files to the list on every call after startup
+    il.adjacent = false;
 
     return first_entry;
 }
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -179,17 +179,7 @@ ImageList::add(const std::vector<std::filesystem::path>& sources)
         sort();
     }
 
-    // disable loading of adjacent files, otherwise fs mon will add unnecessary
-    // files to the list on every call after startup
-    adjacent = false;
-
     return added;
-}
-
-std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path)
-{
-    const std::scoped_lock lock(mutex);
-    return add(path, true);
 }
 
 ImageEntryPtr ImageList::remove(const ImageEntryPtr& entry, const bool forward)
@@ -412,6 +402,9 @@ std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path,
         }
 
         if (!adjacent) {
+            if (fsmon) {
+                FsMonitor::self().add(abs_path);
+            }
             return added;
         }
         abs_path = abs_path.parent_path();
@@ -490,9 +483,6 @@ ImageEntryPtr ImageList::add_file(const std::filesystem::path& path,
     entry->size = std::filesystem::file_size(path);
     if (!add_entry(entry, ordered)) {
         return nullptr;
-    }
-    if (fsmon) {
-        FsMonitor::self().add(path);
     }
     return entry;
 }
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -159,12 +159,12 @@ ImageEntryPtr ImageList::load(const std::filesystem::path& list_file)
     }
     file.close();
 
-    std::list<ImageEntryPtr> added = load(sources);
+    std::list<ImageEntryPtr> added = add(sources);
     return added.empty() ? nullptr : added.front();
 }
 
 std::list<ImageEntryPtr>
-ImageList::load(const std::vector<std::filesystem::path>& sources)
+ImageList::add(const std::vector<std::filesystem::path>& sources)
 {
     // preserve order while inserting only if it can avoid sorting entire list
     const bool add_ordered = sources.size() == 1;
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -141,53 +141,6 @@ ImageList& ImageList::self()
     return singleton;
 }
 
-ImageEntryPtr ImageList::load(const std::vector<std::filesystem::path>& sources)
-{
-    if (sources.empty()) {
-        return nullptr;
-    }
-
-    {
-        const std::scoped_lock lock(mutex);
-        for (const auto& it : sources) {
-            add(it, false);
-        }
-        sort();
-    }
-
-    // disable loading of adjacent files, otherwise fs mon will add unnecessary
-    // files to the list
-    adjacent = false;
-
-    // get first entry of the source list (not of the contents of the sources)
-    ImageEntryPtr first = nullptr;
-    for (const auto& path : sources) {
-        first = find(path);
-        if (first) {
-            break;
-        }
-        if (std::filesystem::is_directory(path)) {
-            const std::filesystem::path abs_path =
-                std::filesystem::absolute(path).lexically_normal();
-            auto it =
-                std::find_if(entries_arr.begin(), entries_arr.end(),
-                             [&abs_path](const ImageEntryPtr& entry) {
-                                 return abs_path == entry->path.parent_path();
-                             });
-            if (it != entries_arr.end()) {
-                first = *it;
-                break;
-            }
-        }
-    }
-
-    if (!first && !entries_arr.empty()) {
-        first = entries_arr.front();
-    }
-
-    return first;
-}
-
 ImageEntryPtr ImageList::load(const std::filesystem::path& list_file)
 {
     std::ifstream file(list_file);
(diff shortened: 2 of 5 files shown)


src/imagelist.cpp
```diff
@@ -206,32 +161,145 @@ ImageEntryPtr ImageList::load(const std::filesystem::path& list_file)
...
+    {
+        const std::scoped_lock lock(mutex);
+        for (auto& path : sources) {
+            if (ImageEntry::is_special(path)) {
+                const ImageEntryPtr entry =
+                    add_special_source(path, sources.size() == 1);
+                if (entry) {
+                    added.emplace_back(entry);
+                    if (first == added.end()) {
+                        first = --added.end();
```
artemsen: How much do you love writing complex and incomprehensible code! =)
This is some kind of creepy multi-stage logic with a game of iterators.
Maybe someday I'll get around to it and try to understand it all, but either you simplify it, or I won't be back here anytime soon, sorry =)

src/viewer.cpp
```diff
@@ -9,6 +9,7 @@
 #include "log.hpp"
 #include "render.hpp"
 #include "resources.hpp"
+#include "src/imagelist.hpp"
```
artemsen: already included in viewer.hpp

src/imagelist.cpp
```diff
@@ -9,9 +9,11 @@
 
 #include <algorithm>
 #include <cassert>
+#include <filesystem>
```
artemsen: already included in hpp

src/imagelist.cpp
litoj: Well, the problem is that you like complex logic for deciding which picture should be displayed first.

The simplest and most logical answer is, that it should be the first one loaded from the sources according to the sorting order. But you for some reason think that it is better to return an opened file, even if it was at the end of a list of directories that didn't have any direct images (`{/, file, .}` will open `file` instead of sth in `/`), but you _would_ display the directory image first if it was a direct child of the directory and not of a subdirectory.

Now _that_ is complex behaviour if you ask me. If you were willing to skip this unnecessary step, everything would get much easier.

artemsen: > Now _that_ is complex behaviour if you ask me. If you were willing to skip this unnecessary step, everything would get much easier.

You apparently still don't understand why this is necessary. This behavior is caused by the "adjacent" flag. When it is enabled, an attempt to open a specific image from a directory with other files will result in the display of a completely different file than the user expects.
Therefore, there is a distinction between load and add.
You've mixed everything together and not done it in the best possible way.
You only need to find the first file the first time you launch the application, not every time you add something to the list.

litoj: My initial motivation for joining the `load` and `add` was to make the behaviour of `viewer.open` the same as the `load` in order to not confuse the user (-;. However, there is no reason in general for why these two should be any different - they use the same underlying logic and both can accept the exact same types of input (special entry, file, directory).

> You apparently still don't understand why this is necessary. 

Well, perhaps if that weird piece of code had any comments to describe why it is there, it would have been clearer =). The reason you give is signifficantly different from the behaviour that is implemented.

If all that is the goal is to determine the first file added by the user then there are only 2 scenarios - the first argument/input is an entry (special or file), or it is a directory and the first entry of the directory should be returned.

If we add the file before the PWD, then the `adjacent` value doesn't matter because the file will always precede the directory contents.

So the only case when you'd want to look beyond the first added entry would be when a directory is added. In that case we can just find the sorted minimum of the added files (just like I am implementing it) and make it precede the rest of the directory contents. And with that, it would be always possible to just return the first element and job done.

I will modify it to work this way and we will see.

src/imagelist.cpp
```diff
@@ -206,32 +160,141 @@ ImageEntryPtr ImageList::load(const std::filesystem::path& list_file)
...
+        added.splice(added.end(), dir_entries);
+    }
+
+    // sort only if multiple entries could have been added
+    if (sources.size() > 1 || std::filesystem::is_directory(sources.front()) ||
+        (adjacent && !ImageEntry::is_special(sources.front()))) {
+        sort();
+    }
+
+    // disable loading of adjacent files, otherwise fs mon will add unnecessary
```
artemsen: When this was done at launch, it seemed logical to disable it immediately after startup. Now it's disabled on every add operation, creating the illusion that it might be activated somewhere at runtime.

src/imagelist.cpp
artemsen: > If we add the file before the PWD, then the `adjacent` value doesn't matter because the file will always precede the directory contents.

If by PWD you mean the current directory, then you have misunderstood this flag again.
We need to process not the current working directory (CWD), but the directory where the file is located, so `swayimg ../photo.jpg` adds all files from the parent directory.

litoj: PWD stands for parent working directory. You can see that the code clearly does just what it should and can verify that it works as expected. Why bring up something unrelated that makes it seem like there is a confusion despite the code clearly showing everything is understood well?

src/imagelist.cpp
litoj: Yeah. Tbh given that it could be useful to be able to use the adjacent flag even after init. That would mean that it should be a method parameter, in which case the initial list population should be also done via add() and therefore via lua and therefore lua should get the list of cli args (limited to the file list).

Though, as I remember, you weren't particularly thrilled about this idea when I brought it up last time, so for keeping the current behaviour it would suffice to disable the adjacent flag via a call in application load instead of the imagelist method.

artemsen: > call in application load instead of the imagelist method

You are spreading the logic of the list across other components, which is not very good.

src/imagelist.cpp
```diff
@@ -206,32 +160,139 @@ ImageEntryPtr ImageList::load(const std::filesystem::path& list_file)
     }
     file.close();
 
-    return load(sources);
+    std::list<ImageEntryPtr> added = add(sources);
+    return added.empty() ? nullptr : added.front();
 }
 
-std::list<ImageEntryPtr> ImageList::add(const std::filesystem::path& path)
+// NOLINTBEGIN(readability-function-cognitive-complexity)
```
artemsen: I specifically added a readability analyzer so you wouldn't think I was just being picky. But you just turned it off. It's kind of funny =)

litoj: No, what's kind of funny is that was your approach in your commit as well for all but one cases running up against this useless warning that requires you to split logic into more methods even though they will be used just once, meaning the logic splitting just causes more confusion because you have to jump around between places to find out what is actually going on instead of simply reading the code as it is executed.

All I can do is split the body of the _for_ cycle back into the private `add`. Which would have just that one use and the logic would be hard-coupled with the `load` method, in turn rendering the code separation into another method pointless because you cannot reuse the code anyway. The only benefit you get from it is one level less of indentation.

artemsen: > is that was your approach in your commit as well

Yes, this is indeed true and I am ready to answer for every such case:
- BMP decoder is a very old and complex code;
- JP2 decoder: YUV to RGB conversion function is copied from upstreamx;
- PNM decoder: not my code at all, but I have refactored it already;
- LUA engine: there are a lot of template bindings, so using some kind of fabric just makes it more complex.

I can't fix everything at once, but I will. And it doesn't mean we can ignore analyzer warnings in new code.

src/imagelist.cpp
litoj: I see it as putting the logic where it belongs - imagelist takes care of changes to the list (the domain/data level, if you will). Application takes care of the behaviour (business logic). Therefore the logic that makes the loading different the first time from all the times after belongs to application and not imagelist.

Wouldn't you agree?

src/application.cpp
```diff
@@ -231,20 +228,26 @@ ImageEntryPtr Application::il_initialize()
...
+            il.add(std::vector<std::filesystem::path> { "." }).front();
     }
 
     if (Log::verbose_enable()) {
         Log::verbose("Image list loaded in {:.6f} sec", timer.time());
     }
 
+    // disable loading of adjacent files, otherwise fs mon will add unnecessary
+    // files to the list on every call after startup
+    il.adjacent = false;
```
artemsen, line 249: Let me reiterate, the internal logic of an image list shouldn't be controlled by external components. I don't know how to explain this more clearly, but if you try to cover this functionality with unit tests, you'll realize that testing the list and the application together is not very convenient, and refactoring such code is problematic.

src/imagelist.cpp
artemsen: > Wouldn't you agree?

No. Adjacent relates to image list and not controlled by the Application at all.

src/imagelist.cpp
litoj: No problem, sorry for the hassle. I was just going through the commit and thought it was the default approach, so I just copied what I saw.

Now, after the nudge from you that I needed, I realized I can easily just adjust the private add() method with very minimal compromises, which I have done.

src/imagelist.cpp
litoj: > Adjacent relates to image list and not controlled by the Application at all.

Except that it adjusts the behaviour, which the user can do, and which the imagelist doesn't do anywhere else. This is clearly business logic and imagelist exposes these settings for a reason. Imagelist itself doesn't change any of its settings, the only modifications it does on its own is managing the entries - _nothing_ else.

src/application.cpp
litoj: Well in that case I have to explain why this belongs where I put it and not where it was. In fact it fixed a bug that your approach would enforce.

Loading is done in `application`. For the duration of the entry loading process the `adjacent` flag is expected to be used. This means that _all_ calls to `load` or `add` are expected to add also the parent directory the file resides in, regardless of how many times the `load`/`add` gets called.

In your approach, if there is a `from_file` param, then the flag will be disabled and won't work for the second `load()` call that might occur when `sources` are also specified.

This is not only _the_ correct approach, but also the _only_ correct approach.

artemsen: > In your approach, if there is a `from_file` param, then the flag will be disabled and won't work for the second `load()` call that might occur when `sources` are also specified.

Good catch!
I still haven't come to a clear opinion on how to handle the situation when there is both from-file and a list from the command line. The situation is especially strange if both from-file and stdin are specified at the same time.
It's probably worth giving priority to from-file, then there won't be a problem with dual loading.

litoj: Why are you still opposed to my approach, even though it perfectly suits the situation and given how much more logic there is around adding the initial files it would only make sense to also add this line?

I understand that you believe it is imagelist logic that is being changed, but this logic is behaviour specific to startup, and imagelist is supposed to handle just adding and removing files. There is no reason imagelist should need to differentiate between adding initial files and just adding files.

That is my primary reason for believing that the flag belongs to `application` - because that is a change of how the files are loaded.

Imagelist has a setting for it, and if it changed its own setting, then it would seem weird to make a setting that it can overwrite. This way, it stays a fixed setting that the caller may change, but is guaranteed to stay the way the last caller has set it to.

artemsen:
Okay, you've convinced me. Let's give it a try.

artemsen:
This is the third time the program has crashed after your PR. Please remember to test with an empty list.

artemsen:
Also gallery stopped working =) May be it is from #506.

litoj:
I will fix it

artemsen:
> I will fix it

I have fixed it already and reverted 506. please, test before PR.

litoj:
Well, you see, I _do_ test the changes manually, but only _before_ the PR. So your condition would be actually satisfied. After creating the PR, I only run the meson tests because I usualy make just minor changes that you request.

The problem is that I expect that when I rebase and nothing needs merging my changes should work as well. However, since bd9dbb1 This is not the case and I cannot even start the app because of text rendering errors:

```
==40861==The signal is caused by a READ memory access.
==40861==Hint: address points to the zero page.
    #0 0x558f2c11197b in Font::render(std::__cxx11::basic_string<char, std::char_traits<char>, std::all
ocator<char> > const&) ../src/font.cpp:228
    #1 0x558f2c24c3c9 in Text::Line::update(Font&, std::map<std::__cxx11::basic_string<char, std::char_
traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::all
ocator<char> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>
 > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<
char> > const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > > co
nst&) ../src/text.cpp:467
    #2 0x558f2c249500 in Text::update() ../src/text.cpp:256
    #3 0x558f2c26b2d3 in Viewer::update_text(Viewer::TextUpdate) const ../src/viewer.cpp:755
    #4 0x558f2c26573f in Viewer::set_scale(double, Point const&) ../src/viewer.cpp:290
    #5 0x558f2c2651a6 in Viewer::set_scale(Viewer::Scale) ../src/viewer.cpp:260
    #6 0x558f2c26a58b in Viewer::set_image(std::shared_ptr<Image> const&) ../src/viewer.cpp:702
    #7 0x558f2c268130 in Viewer::set_current(std::shared_ptr<ImageEntry> const&) ../src/viewer.cpp:521
    #8 0x558f2c269aab in Viewer::open(std::shared_ptr<ImageEntry> const&, bool) ../src/viewer.cpp:644
    #9 0x558f2c2675fd in Viewer::activate(std::shared_ptr<ImageEntry> const&, Size const&) ../src/viewe
r.cpp:466
    #10 0x558f2c07bf1e in Application::run() ../src/application.cpp:88
    #11 0x558f2c3479ff in main ../src/main.cpp:275
    #12 0x7f6829427740  (/usr/lib/libc.so.6+0x27740) (BuildId: 020d6f7c33b2413f4fe10814c4729dce1387f049
)
    #13 0x7f6829427878 in __libc_start_main (/usr/lib/libc.so.6+0x27878) (BuildId: 020d6f7c33b2413f4fe1
0814c4729dce1387f049)
    #14 0x558f2c07b084 in _start (/home/kepis/Documents/personal/PG/swayimg/_debug/swayimg+0x8a084) (Bu
ildId: d7e2bff41df8b23911e561e03928fda799a331f8)

==40861==Register values:
rax = 0x0000000000000000  rbx = 0x00007b6818277580  rcx = 0x0000000000000000  rdx = 0x0000000000000098 
 
rdi = 0x0000000000000000  rsi = 0x000000000000003f  rbp = 0x00007ffddf6d5ea0  rsp = 0x00007ffddf6d5c30 
 
 r8 = 0x00007f68295729c0   r9 = 0x00000000000000f3  r10 = 0x0000000000000000  r11 = 0x00007f682b0e0f00 
 
r12 = 0x00000f6d0304ee80  r13 = 0x00007b6818277400  r14 = 0x00007ffddf6d6100  r15 = 0x00007ffddf6d5cf0 
 
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV ../src/font.cpp:228 in Font::render(std::__cxx11::basic_string<char, st
d::char_traits<char>, std::allocator<char> > const&)
==40861==ABORTING
```

artemsen:
Rebase is not the problem. You called [front()](https://github.com/artemsen/swayimg/commit/d2c011b208d9c19bbdd78371477ee7bcb9b83041) for empty list.

litoj:
Yes, that one was my fault, but I still cannot run the application since that commit. You can see the segfault - it has nothing to do with the issue that I caused later and only in my PR.

artemsen:
> segfault

Fixed. Just a typo.

litoj:
What was the reason for reverting the gallery change? For me it still works perfectly fine.

PR #506, opened 2026-06-03
Simplify redraw and skip unnecessary redrawing

litoj:


src/appmode.cpp
```diff
@@ -65,9 +65,12 @@ void AppMode::handle_imagelist(const ImageListEvent,
     text.update();
 }
 
-bool AppMode::is_active() const
+void AppMode::try_redraw() const
```
artemsen: I don't see any reason for these changes. It even sounds strange: "try to redraw."

src/gallery.cpp
```diff
@@ -209,83 +209,66 @@ void Gallery::reload()
     tpool.cancel();
     tpool.wait();
 
-    // clear cache
-    const std::scoped_lock lock(mutex);
-    cache.clear();
+    {
```
artemsen: Please stop forcing changes that I have [already rejected](https://github.com/artemsen/swayimg/pull/499#discussion_r3328709113).

src/appmode.cpp
litoj: Ofc technically there isn't one, but it makes the code a great deal shorter in places that are so trivial that the check for redrawing is longer than the entire business logic of the method body, which makes it easier to overlook the actually important code.

I've renamed it to `redraw_if_active` and I will drop those changes if you really wish me to. But I strongly believe that removing duplicate code especialy one that is used as heavily as this is a good thing for both readability and maintainability.

src/gallery.cpp
```diff
@@ -616,13 +591,14 @@ void Gallery::load_thumbnail(const ImageEntryPtr& entry)
...
 
     const std::scoped_lock lock(mutex);
-
     if (pm) {
         cache.insert_or_assign(entry, pm);
     }
     active.erase(entry);
 
-    Application::redraw();
+    if (layout.get_scheme().back().img->index >= entry->index) {
```
artemsen: There is no need to externalize the logic of working with indexes.

src/luaengine.cpp
```diff
@@ -958,7 +958,7 @@ void LuaEngine::bind_gallery_api()
 {
     // check if required mode is active
     auto ensure_active = [this](const char* fname) {
-        if (!Gallery::self().is_active()) {
+        if (Application::self().get_mode() != Application::Mode::Gallery) {
```
artemsen: I don't think it's gotten any better.

src/appmode.cpp
artemsen: This is too vague a name.
What's active? An app? A thread?
Let's call it `redraw_only_if_current_mode_is_active()`, make it template function, add couple of callbacks and wrap with macros =)
Seriously, I don't see the point in these changes.

src/gallery.cpp
litoj: I read it as you find it better to get the distance via calling imagelist distance, right?

Ok, will do.

src/appmode.cpp
litoj: I really don't understand your philosophy. You say you want the code to be readable and maintainable, yet this attempt at making the code simpler and more readable as well as the attempt to lock only the code that needs to be locked, which helps with maintainability a lot, seem to be completely outside your interest.

src/gallery.cpp
litoj: Since I've realized deletion redraw is completely useless because the layout doesn't change there at all I excluded it from the redraw check logic. With this increase of logic past the code that needs locking it seems only appropriate for the lock to be restricted only to the code that needs it, so that future additional code doesn't suddenly cause a deadlock because of not realizing that there is an active lock that had no reason to be there.

Locks should always be used only for the time necessary - not more. It helps with both maintainability as well as multithreading performance. I really don't know what's your deal for trying to avoid the correct approach.

litoj: I also don't understand what you dislike about working with indexes. They are used everywhere in gallery and layout for the exact same purpose of determining the valid distance from the active image under various scenarios (`clear_inviisble_thumbnails`, `load_thumbnails` and practically every method that works with thumbnails which is the whole purpose of gallery).

Given that `il.distance` isn't being used in almost any of them I'd assume that there is no reason to be using it here either. It would both boost performance because two locking calls could be avoided and it would be consistent with the rest of the file. Please give a second thought to what it actually is, if anything, that you dislike about this improvement.

artemsen: > They are used everywhere in gallery and layout

They are not used in layout. In the gallery, yes, and it was my fault that I missed it in the review.
Firstly, you're once again spreading the list-handling logic across other components. Secondly, working with indexes requires locks, since they can be modified by another thread.

src/appmode.cpp
artemsen: Apparently we have different views on the concept of "readability".

src/gallery.cpp
```diff
@@ -606,25 +610,23 @@ void Gallery::load_thumbnail(const ImageEntryPtr& entry)
...
-void Gallery::queue_thumbnail(const ImageEntryPtr& entry)
-{
-    if (!get_thumbnail(entry) &&   // not yet loaded
-        !queue.contains(entry) &&  // not queued
-        !active.contains(entry)) { // not currently loading
-        queue.insert(entry);
-        tpool.add([this, entry]() {
-            load_thumbnail(entry);
-        });
+    // redraw only if visible
```
artemsen: Nice idea, but lets' move it to `Layout::is_visible`.

artemsen:
I reverted this changes. They break gallery as turns out.

PR #507, opened 2026-06-08
imagelist: Support bulk remove()

litoj:
Part of the change for aligning the imagelist entry modifying methods to have the same input+output.

### DIFF
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -12,6 +12,7 @@
 #include <fstream>
 #include <mutex>
 #include <random>
+#include <set>
 #include <string>
 
 /**
@@ -180,6 +181,62 @@ ImageList::add(const std::vector<std::filesystem::path>& sources)
     }
 
     return added;
+}
+
+std::list<ImageEntryPtr>
+ImageList::remove(const std::vector<std::filesystem::path>& paths)
+{
+    if (paths.empty() || entries_arr.empty()) {
+        return {};
+    }
+
+    std::list<ImageEntryPtr> removed;
+    std::set<size_t> removed_indexes;
+
+    const std::scoped_lock lock(mutex);
+    for (const auto& path : paths) {
+        auto it = entries_map.find(
+            ImageEntry::is_special(path)
+                ? path
+                : std::filesystem::absolute(path).lexically_normal());
+        if (it != entries_map.end()) {
+            Log::verbose("Remove image entry {}", path.filename().string());
+            it->second->removed = true;
+            removed.emplace_back(it->second);
+            removed_indexes.insert(it->second->index);
+            entries_map.erase(it);
+        }
+    }
+
+    if (removed.size() == 1) {
+        entries_arr.erase(entries_arr.begin() + removed.front()->index);
+        reindex(removed.front()->index);
+    } else if (!removed.empty()) {
+        if (entries_map.empty()) {
+            entries_arr.clear();
+            return removed;
+        }
+
+        // shift and reindex - move the entries directly to their new position
+        removed_indexes.insert(entries_arr.size());
+        auto it = removed_indexes.begin();
+        const auto end = removed_indexes.end();
+        size_t from = *it;
+        size_t to = from;
+
+        while (++it != end) {
+            from = from + 1;
+
+            for (const size_t next = *it; from < next; ++from, ++to) {
+                entries_arr[to] = entries_arr[from];
+                entries_arr[to]->index = to;
+            }
+        }
+        entries_arr.erase(entries_arr.begin() + entries_map.size(),
+                          entries_arr.end());
+    }
+
+    return removed;
 }
 
 ImageEntryPtr ImageList::remove(const ImageEntryPtr& entry, const bool forward)
--- a/src/imagelist.cpp
+++ b/src/imagelist.cpp
@@ -12,7 +12,6 @@
 #include <fstream>
 #include <mutex>
 #include <random>
-#include <set>
 #include <string>
 
 /**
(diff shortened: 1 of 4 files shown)


src/imagelist.hpp
```diff
@@ -65,11 +65,11 @@ class ImageList {
 
     /**
      * Remove image entry from the list.
-     * @param entry entry to remove
-     * @param forward direction of nearest returned entry
-     * @return valid nearest entry or nullptr if list is empty
+     * @param entry_paths entry paths to remove
+     * @return list of removed entries
      */
-    ImageEntryPtr remove(const ImageEntryPtr& entry, const bool forward = true);
```
artemsen, line 66: This is a rather controversial change.
Multiple deletion is a very rare situation that can only happen through a direct call from a lua script.
But removing a specific entry is a fairly common operation when loading files and thumbnails.
And now it will work even slower, since it will have to calculate the hash of the path to find the required entry, which is already known in advance.

litoj: If you wish, I can make it accept `vector/list<IEP>` instead.

All I am sure of is that the bulk remove is absolutely necessary, because doing it one-by-one is incredibly costly because of the constant reindexing.

artemsen: I would create standalone function `void remove(const std::vector<ImageEntryPtr>& entries)`.
We don't have to look for the nearest next entry, so this makes the it easier.

test/imagelist_test.cpp
```diff
@@ -658,3 +658,52 @@ TEST(ImageListTest, MoveFromRemoved)
...
+    size_t i = 0;
+    for (const auto& entry : il.get_all()) {
+        EXPECT_EQ(entry.index, i);
+        EXPECT_EQ(entry.path, paths[i]);
+        ++i;
+    }
+    EXPECT_EQ(i, paths.size());
+}
+
+TEST(ImageListTest, RemoveBulk)
```
artemsen, line 708: It's nice that you started writing new tests!
BTW, from the [actions page](https://github.com/artemsen/swayimg/actions/runs/27131567072/job/80073526131?pr=507) (Upload report) you can download [code coverage report](https://github.com/artemsen/swayimg/actions/runs/27131567072/artifacts/7477700316).

src/imagelist.cpp
```diff
@@ -182,6 +183,62 @@ ImageList::add(const std::vector<std::filesystem::path>& sources)
...
+    const std::scoped_lock lock(mutex);
+    for (const auto& path : paths) {
+        auto it = entries_map.find(
+            ImageEntry::is_special(path)
+                ? path
+                : std::filesystem::absolute(path).lexically_normal());
+        if (it != entries_map.end()) {
+            Log::verbose("Remove image entry {}", path.filename().string());
+            removed.emplace_back(it->second);
+            removed_indexes.insert(it->second->index);
```
artemsen, line 205: You'll be able to traverse the array faster than processing the tree; reindexing is a very fast operation.
No need for these set-related fussies.

src/imagelist.cpp
```diff
@@ -182,6 +183,62 @@ ImageList::add(const std::vector<std::filesystem::path>& sources)
...
+            it->second->removed = true;
+            entries_map.erase(it);
```
artemsen, line 206: When deleting from array, you must also delete it from the map. There is no need to put this off until later.

litoj: What am I delaying? I need to do everything with the iterator first, because `map.erase` invalidates it.

artemsen: I mean that map and array should be changed together. You remove from map here, but array latter.

src/imagelist.cpp
litoj: Good point.

litoj: Wait, I take it back.

The reason I am doing it this way, is so that the moving of all entries in the entries_arr is also minimized.

Yes, the reindexing is fast, but shifting all the entries in the actuall array not so much.

We're talking O(mlogm) (m=size(removed)) vs O(mn) (n=size(entries))

artemsen: This won't be any faster than your nested-loop games. Especially considering that erase doesn't return the vector memory back to the system, and the savings from calling it only once are highly questionable.

litoj: Just to clarify, you are saying that this:
```
arr=a|b|c|d - a-c -> m×O(n) shifts

arr=b|c|d
arr=c|d
arr=d
```
is faster than:
```
arr=a|b|c|d - a-c -> m×O(logm) + O(n)
set=a
set=a|b
set=a|b|c
arr=d     (O(n) shift)
```

suggesting that `logm` is higher than `n`. You know that's impossible beyond the case of n=1, right?

artemsen: 1. Why do you only consider the ideal case? Who said that only a consecutive block will always be removed?
2. I'd prefer a simple code to dubious savings, especially since yours is clearly a bit crazy (sorry, no offense).

litoj: my code isn't crazy, it is unfortunatelly just how the standard shifting works - you just take one index and move it to the other index and the gap between the new and old indexes increases as you come acros more and more positions that were removed.

litoj: You're right that it could also be:
```
arr=a|b|c|d - c-a -> m×1

arr=a|b|d
arr=a|d
arr=d
```
Which would also work better than constructing the whole set first. On the other hand, that is also an extreme case. My approach is mathematically much better on average and I can provide benchmark results if that would give me a chance of convincing you =)

artemsen: > standard shifting works

I understand that. But I'd still prefer a simple code. People come and go, but this code remains under my support.

PR #508, opened 2026-06-08
gallery: Avoid unnecessary redrawing

litoj:
in #506 the problem was that I was using `il.distance()` as a `cmp` method, but its value is actually the opposite of that -> changing `<=` to `>=` is all that was needed

### DIFF
--- a/src/gallery.cpp
+++ b/src/gallery.cpp
@@ -406,6 +406,7 @@ void Gallery::handle_imagelist(const ImageListEvent event,
                     return;
                 }
                 switch_current();
+                break;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


artemsen:
> in #506 the problem was that I was using `il.distance()` as a `cmp` method, but its value is actually the opposite of that -> changing `<=` to `>=` is all that was needed

I looked closely at this logic of the previous version and thought that I must have misunderstood something, since it works for you.
I won't blindly trust anymore, sorry =)

litoj:
> I won't blindly trust anymore, sorry =)

Yeah, sorry. I should have tested manually again. I will try to remember that next time.

PR #510, opened 2026-06-09
Support empty image list and no selection

litoj:
In order to avoid crashes in situations where the user happens to remove all entries, but the continues working in lua interacting with the image list (for example by adding some images back), the modes need to be able to have a null selected.

This is already the state at initialization, but it wasn't expected by the modes while active. So for the most part this was just making it accept an empty selected entry also while being active and during rendering (that's why a dummy ImagePtr was introduced).

In order to not go over the "complexity" check because of the simple `if(!image->entry) return;` I had to rewrite the thumbnail scheduler.

The `imagelist::get` can now resolve nullptr in combination with any direction.

Edit: sorry for the repeated commits. I had a good idea, but pushed it half-done, so I had to push again to fix it properly.

### DIFF
--- a/src/imagelist.hpp
+++ b/src/imagelist.hpp
@@ -27,14 +27,20 @@ public:
 
     /** Direction of the next entry. */
     enum class Dir : uint8_t {
-        First,      ///< First entry in the list
         Last,       ///< Last entry in the list
+        First,      ///< First entry in the list
         Next,       ///< Next entry
         Prev,       ///< Previous entry
         NextParent, ///< Next entry with different parent
         PrevParent, ///< Previous entry with different parent
         Random      ///< Random entry
     };
+
+    constexpr static bool is_forward(const Dir dir)
+    {
+        return dir == Dir::Last || dir == Dir::Next || dir == Dir::NextParent ||
+            dir == Dir::Random;
+    }
 
     /**
      * Get global instance of image list.
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -816,14 +816,13 @@ This function adds a file to the image list and then opens it in the viewer.
 ### swayimg.viewer.get_image
 
 ```lua
-swayimg.viewer.get_image() -> swayimg.image
+swayimg.viewer.get_image()
 ```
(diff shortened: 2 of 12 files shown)


src/imagelist.hpp
```diff
@@ -27,15 +27,20 @@ class ImageList {
...
         Next,       ///< Next entry
         Prev,       ///< Previous entry
         NextParent, ///< Next entry with different parent
         PrevParent, ///< Previous entry with different parent
         Random      ///< Random entry
     };
 
+    constexpr static bool is_forward(const Dir dir)
+    {
+        return !(static_cast<uint8_t>(dir) % 2);
```
artemsen: Wow! So disgusting it's even funny.
But seriously, after lines like these, I lose all desire to analyze the rest of the code.
I will take your ideas into account in this PR, but it is easier for me to do it myself than to explain “why it can’t be done this way.”

litoj: Well, I suppose you're right I am somewhat abusing the internal enum representation. But other than this (which I corrected to your desires already), I do believe you won't have anything else to be disgusted by in this PR.

litoj:
I think you'd benefit if I was to give a peer review as well before making larger changes. Giving it a little while to think about any changes for a bit longer can do the project only good. And after all, we're all human, so we all can benefit from a peer review - not just outsiders.

In dc35bba half of the places changes are `if (!image)` and half `if (image)`, apparently based on number of lines. If you wish to not degrade the readability and maintainability you should stick to one approach, idealy the more explicit one (`if (!image) return`) so that it is easier to spot if the special case is being handled properly or not. (that's for example the reason I was pushing the separating mutex block in gallery)

As it stands, as soon as some changes are needed to one of the `if (image)` it will suddenly become too long and will have to be changed to a separate return statement.

I can also see the API hasn't been updated, nor have the text image reset methods. It seems pretty unlikely to me that all cases of the text image resets have been taken care of to avoid the necessity to update `text.cpp` as well. But the lua api definitely needs to reflect the changes.

PR #511, opened 2026-06-09
gallery: Separate thumb scheduling and rendering

litoj:
Refreshes thumbnail queue only when meaningful changes have been done - switching entry or modifying the layout by other means.

Given how quick redrawing already was, there is no benefit in having these heavier operations in `redraw` because every update corresponds to 1 redraw and none get skipped unless the main loop is blocked by multiple changes caused by the lua code.

This makes for the following comparison:

When recalculating on every redraw:
- `-`: recalculates even when unrelated changes have been made (color change, text size, etc.)
- `+`: recalculates only once when user calls `gallery.select()` multiple times in a row (probably trying to select a specific image -> set_current should be exposed to the api instead)

The opposite is the case for my approach.

I found a fun quirk for why the layout tests were not covering the `col<0` and `col > max` cases. I am not making a test to force that scenario though. It is much easier to adjust the implementation code, which I had to do anyway because of the update skipping.

### DIFF
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -21,7 +21,11 @@ void Layout::update()
     if (!sel_entry) {
         sel_entry = first_entry;
     }
-    assert(!sel_entry->removed);
+    if (sel_entry->removed && !select(Right) && !select(Left)) {
+        sel_entry = nullptr;
+        scheme.clear();
+        return;
+    }
 
     columns = std::max(static_cast<size_t>(1),
                        window.width / (thumb_size + thumb_padding));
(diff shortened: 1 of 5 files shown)


src/layout.cpp
```diff
@@ -231,6 +239,11 @@ ImageEntryPtr Layout::get_selected() const
 
 bool Layout::is_visible(const ImageEntryPtr& entry) const
 {
+    if (scheme.empty() || !entry || scheme.front().img->removed ||
```
artemsen: What a strange logic. Why only the first and the last?

litoj: This is an example of logic that I hate even myself. The situation this resolves is as follows:

Files can become deleted only when in `handle_imagelist`. In there, when the selected file is deleted, the check in `layout.update` fails so I tried to update the `sel_entry` first. However, because of the optimization that skips the updating if the new selected entry is already visible, I am calling this method, and `il.distance()` requires both arguments to not be null and not be removed.

I have now simplified the hack to `select(dir)` as just checking `sel_entry->removed`, but other suggestions are also welcome.

litoj: Ok, couple hours later and I have exhausted my ideas. I tried the forementioned approach, but that doesn't cover the case when entries have been removed, but `handle_imagelist > layout.update` hasn't been reached yet and the `load_thumbnail` tries to test if a redraw is necessary.

Therefore I resorted to just explaining the purpose of those weird lines. But if you have a better idea that is not just _skip the optimizing_ :sweat_smile: then I'd be happy to try it.

PR #518, opened 2026-06-21
Fix stopping key repeat on old/unrelated key release

Krzysztof-WW:
Hi, 
so there is an issue that auto repeat stops when any key is released even if that key is not repeated any more. It causes problems for example in gallery mode when quickly changing direction, it just stops switching image even if one arrow is pressed.


artemsen:
Thanks!

PR #532, opened 2026-07-15
Preserve `adjacent` setting after initialization.

de-vri-es:
Instead, temporarily disable it when processing fsmon events.

This ensures that dragging and dropping images will still honor the `adjacent` setting.

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -250,10 +250,6 @@ ImageEntryPtr Application::il_initialize() const
         Log::verbose("Image list loaded in {:.6f} sec", timer.time());
     }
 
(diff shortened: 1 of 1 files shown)


artemsen:
Sorry, I don't like that logic.
I don't think this flag should affect the drag-and-drop operation.

de-vri-es:
Why not? Now, if you drag-and-drop a file, you end up with a single image in the list. This is different from opening the same image directly, which seems weird to me.

artemsen:
> which seems weird to me

It's wired for me when I drag 3 images but got 300 =)
But... Why Not?

PR #542, opened 2026-07-27
lua: Refactor of api declaration order

### LINKED ISSUE #541: Docs request: consistent order of attribute declaration&definition
Currently, the order of assigning the default values of variables for all parts of the api in .cpp is different from the order of the lua api documentation and the lua api implementation (luaengine.cpp).

This makes it harder to check if all parts are the same and what has changed etc. Which is especially unpleasant when trying to maintain a wrapper of the api.

Would it be welcome if I rearranged the luaengine code and the lua docs?

litoj:
introduce consistency with module hpp sources

Resolves #541 

Edit: That sign-off check always catches me unprepared

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -34,15 +34,15 @@ or in the file `/usr/share/swayimg/example.lua` after installing the program.
 ## List of available functions
 
 * General functionality
-  * [swayimg.mode](#swayimgmode): Application mode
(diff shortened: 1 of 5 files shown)


litoj:
Given that gallery thumb, border and padding sizes are all related and play a role in the layout, I'd reorder the implementation code as well to respect this relation.

PR #548, opened 2026-07-30
Fix possible NULL dereference in error messages

### LINKED ISSUE #547: Segmentation fault when failing to load a font
**Describe the bug**
in font.cpp, FT_Error_String is used to show an error message when a font fails to load, but it turns out [it can be a null pointer](https://freetype.org/freetype2/docs/reference/ft2-error_enumerations.html#ft_error_string) depending on how freetype was compiled, so this can lead to a segmentation fault

**To Reproduce**
`echo 'AAEAAAA' | base64 -d > problem; swayimg problem` (swayimg thinks that it's a font for some reason) (i found out about this when i ran swayimg -g in a folder with an aria2 control file)

**Version:**
`5.5-3-gd99aec1`

Exonorid:
Fixes #547 by adding a fallback error message whenever FT_Error_String would return NULL.

### DIFF
--- a/src/font.cpp
+++ b/src/font.cpp
@@ -18,6 +18,18 @@ constexpr const char FALLBACK_CHR = '?'; // character used for absent glyphs
 constexpr size_t POINT_FACTOR = 64;  // default points per pixel (26.6 format)
 constexpr size_t MAX_TEXT_LEN = 120; // max length of text line (characters)
 
+// Include a fallback message when FT_Error_String returns NULL
+// since it can cause a segfault
+const char* FT_Error_String_safe(FT_Error error_code)
+{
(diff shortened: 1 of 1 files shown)


src/font.cpp
```diff
@@ -18,6 +18,18 @@ constexpr const char FALLBACK_CHR = '?'; // character used for absent glyphs
...
+// since it can cause a segfault
+const char* FT_Error_String_safe(FT_Error error_code)
+{
+    const char* error_string = FT_Error_String(error_code);
+    if (!error_string) {
+        error_string = "unknown error";
+    }
+    return error_string;
+}
+#define FT_Error_String FT_Error_String_safe
```
artemsen, line 31: Sorry, but this is a dirty hack =)

PR #551, opened 2026-07-31
Fix macOS build

barracuda156:
Allow building on macOS

### DIFF
--- a/src/ui_wayland.cpp
+++ b/src/ui_wayland.cpp
@@ -17,6 +17,21 @@
 #include <cstring>
 #include <format>
 
+#ifdef __APPLE__
+// Darwin has no pipe2(). Emulate the O_CLOEXEC part; the window
+// between pipe() and fcntl() is benign here, nothing execs
+// concurrently with the drag-and-drop receive path.
+static int pipe2(int fds[2], int /* flags */)
+{
+    if (pipe(fds) == -1) {
+        return -1;
+    }
+    fcntl(fds[0], F_SETFD, FD_CLOEXEC);
+    fcntl(fds[1], F_SETFD, FD_CLOEXEC);
+    return 0;
+}
+#endif
+
 // MIME type for drag-and-drop of raw path
 static constexpr const char* MIME_TEXT_PLAIN = "text/plain";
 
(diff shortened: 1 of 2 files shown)


src/ui_wayland.cpp
```diff
@@ -17,6 +17,21 @@
 #include <cstring>
 #include <format>
 
+#ifdef __APPLE__
```
artemsen, line 20: Swayimg is constantly evolving, and since I'm not a macOS user, I can't guarantee that these patches will work anytime soon. I can't even compile them.
In other words, sooner or later it will turn into dead code.

I'm not against adding macOS support, but first you need to add CI for macOS.

barracuda156: I am not asking to give any guarantees everything continues to work (it is just never the case, regressions happen even in gcc and cmake). However if I add CI for specific macOS versions, which will be different from what I use, I will have to maintain those as well, and dealing with clangs is not something I am keen on. For example, I could not easily build this on 10.15 (thanks, Apple libc++), while the build works on 10.6 (GNU libstdc++).

wayland2

P. S. The build also needs a fix for Big-endian platforms, otherwise colors are totally broken. But that one is independent of macOS.

artemsen: In that case, you'd be better off developing your own fork.

> fix for Big-endian platforms

Yeah, but it is a very rarely systems for now. I don't have any HW to check and support this.

barracuda156: > I don't have any HW to check and support this.

Open-source works pretty well in decentralized manner. Nobody has all hardware, all expertise and all time to maintain everything.

artemsen:
> Open-source works pretty well in decentralized manner. Nobody has all hardware, all expertise and all time to maintain everything.

So, you can fork the project and do whatever you want. But it doesn't mean that you can commit unmaintained code to the master.
Moreover, swayimg doesn't support BE and fixing it is a really big work, and given that I won't be able to check this manually for every release, this will require full test coverage.
Sorry, but I am not ready to accept these changes, especially since they do not completely solve all the problems for working in your OS.

PR #571, opened 2026-09-07
viewer: zoom relative to image size

AminurAlam:
before this the zooming used to add a fixed size to the image which caused small scaled images to change too much and big scaled images to hardly change

- now the image will scale relative to its own size
- pinch_factor had to be inverted for it to make sense
- pinch factor between 5.0 to 0.5 were working fine
- pinching in gallery mode also looks good
- i couldnt figure out how to run tests

left is my change, right is the original behaviour. both done with default config

https://github.com/user-attachments/assets/84c5df0e-3ce4-4a97-af3f-d3c33b13a9d9

https://github.com/user-attachments/assets/716c4d7f-87b1-4f42-9505-0f1af5e60b1b

### DIFF
--- a/src/viewer.cpp
+++ b/src/viewer.cpp
@@ -534,7 +534,7 @@ void Viewer::handle_mmove(const InputMouse& input, const Point&,
 
 void Viewer::handle_pinch(const double scale_delta)
 {
-    set_scale(scale + scale_delta * pinch_factor);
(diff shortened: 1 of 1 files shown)


artemsen:
> * i couldnt figure out how to run tests

There is no test for pinch opertaions.
To run unit tests locally you need to enable them on build (`meson -D tests=enabled`).
Also you can enable github actions for your fork, so you will be able to run CI in your own repo.

PR #573, opened 2026-09-09
wayland: Use axis_value120 for wheel scroll direction

kaosuran:
Hi, started having issues with swayimg at some point with zooming out using a mouse wheel, with the help of claude I think it's fixed now, tested it locally and the zooming works correctly for up/down now, maybe it'll come in handy!

swayimg drives raw wl_pointer bound at version 4 and derived the scroll direction from the sign of every wl_pointer.axis event. That value is a continuous quantity: it is delivered as 0 to notify a v4 client that scrolling stopped, and on a slow high-resolution wheel a single detent arrives as a short burst of events whose signs alternate around zero. Acting on each event individually makes the wheel scroll -- and thus zoom -- in a seemingly random direction; slow scrolling is affected the most. Reproducible on Hyprland with a hi-res wheel (e.g. Logitech G502).

Bind wl_seat at version 8 and take the wheel direction from wl_pointer.axis_value120, which is guaranteed non-zero and reports 120 units per detent. Accumulate axis/axis_value120 within a frame and dispatch in wl_pointer.frame, keeping the sub-detent remainder for the next frame. The legacy axis value is still used as a fallback for touchpads and pre-v8 compositors.

Binding v8 requires non-null listener entries up to axis_value120, so add the (mostly empty) axis_source/axis_stop/axis_discrete/ axis_relative_direction handlers.

### DIFF
--- a/src/ui_wayland.cpp
+++ b/src/ui_wayland.cpp
@@ -178,13 +178,9 @@ public:
         }
     }
 
-    static void on_pointer_axis(void* data, struct wl_pointer*, uint32_t,
-                                uint32_t axis, wl_fixed_t value)
+    static void send_scroll(const UiWayland* ui, uint32_t axis, bool incr)
     {
-        const UiWayland* ui = reinterpret_cast<UiWayland*>(data);
-
         InputMouse::mouse_btn_t btn;
-        const bool incr = value > 0;
         if (axis == WL_POINTER_AXIS_HORIZONTAL_SCROLL) {
             btn = incr ? InputMouse::SCROLL_RIGHT : InputMouse::SCROLL_LEFT;
         } else {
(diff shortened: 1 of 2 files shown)


litoj:
I think touchpads could benefit from this as well. I cannot use my touchpad because the scrolling is so fast I always scroll to the first or last image in gallery in under half a second. This is something I was trying to hack around through lua, but having it right in would be amazing.

artemsen:
It looks like this implementation breaks the legacy scrolling behavior.
However, I liked the idea of ​​using `value120`, so I tweaked your PR a bit.
In addition to supporting hi-res devices, I added a Lua handler that allows for scrolling speed customization.
Thank you!

PR #576, opened 2026-09-10
text: Fixed text not rendering on empty image list

### LINKED ISSUE #575: Text layer doesn't render when imagelist is empty
When imagelist is empty, the text layer doesn't render, only the status does.

litoj:
Resolves #575

### DIFF
--- a/src/appmode.cpp
+++ b/src/appmode.cpp
@@ -160,8 +160,14 @@ void AppMode::switch_current()
     Text& text = Text::self();
     if (entry) {
         text.reset(entry);
-    } else {
+    } else if (text.get_field(Text::FIELD_LIST_INDEX) != "0") {
+        // clear only once to avoid redrawing the status message on every change
         text.clear();
+        text.set_field(Text::FIELD_LIST_INDEX, "0");
+        text.set_field(Text::FIELD_LIST_TOTAL, "0");
+        text.set_field(Text::FIELD_FRAME_INDEX, "0");
+        text.set_field(Text::FIELD_FRAME_TOTAL, "0");
+        text.update();
         text.set_status("Image list is empty");
     }
 
--- a/src/text.cpp
+++ b/src/text.cpp
@@ -248,6 +248,12 @@ void Text::set_field(const std::string& field, const std::string& value)
     }
 }
 
+std::string Text::get_field(const std::string& field) const
+{
+    const auto it = fields.find(field);
+    return it != fields.end() ? it->second : std::string();
+}
+
 void Text::update()
 {
     for (auto& block : blocks) {
--- a/src/text.hpp
+++ b/src/text.hpp
@@ -173,6 +173,13 @@ public:
     void set_field(const std::string& field, const std::string& value);
 
     /**
+     * Get field value.
+     * @param field field name
+     * @return field value or empty string if the field is not set
+     */
+    [[nodiscard]] std::string get_field(const std::string& field) const;
+
+    /**
      * Update text blocks.
      */
     void update();


artemsen:
Did you check this patch with the default config?
image

litoj:
I sure haven't. But how is it possible that it says frame 1/1 even though text.clear was called? Does it not actually reset all the fields?

artemsen:
> But how is it possible that it says frame 1/1 even though text.clear was called?

It **_was not_** called. Check your code.

litoj:
I know, I realized that a minute after writing the comment :D. Should be fixed now

artemsen:
image

I see no reason to accept this patch. Why are these extra handlers needed?

litoj:
what do you mean by extra handlers? If you mean setting the index to 0, I have no problem with dropping that, but I need the textlayer to be able to keep displaying.

You added support for no image display, but the app is still unusable in that state unless we can see the text that we have set it to. It is also the only way to distinguish if we are in gallery or viewer mode, since the displayed icon is the same for both.

artemsen:
> what do you mean by extra handlers?

Now we need to handle the empty fields that are usually populated; for example, `size` is now displayed as `x`.

> app is still unusable in that state unless we can see the text that we have set

And what do you expect from an image viewer if it doesn't have an image to display?

> in gallery or viewer mode

What difference does it make, if there’s nothing to show anyway?

litoj:
> What difference does it make, if there’s nothing to show anyway?

The difference is that the user has keybindings that are different for each mode.

> And what do you expect from an image viewer if it doesn't have an image to display?

That perhaps the user has bookmarks or some other form of adding images back to swayimg. But I cannot orientate myself when I see nothing on the screen.

>Now we need to handle the empty fields that are usually populated; for example, size is now displayed as x.

Do we? If you were fine with a broken behaviour before, why would this differently-broken behaviour be an issue?

litoj:
My goal is only to make no-image display work the same way as the app works under any other circumstances. I see no reason to compromise on functionality because of a separate component.

To me, the UI possibilities even with just the 4 locations are great, and I want t be able to keep my UI even when I am going through a state when no images are visible.

If you were to implement an image filter (as someone requested and where I am encountering this problem), you'd likely come to the conclusion as well that it is quite annoying when the text you are writing just suddenly disappears only because your filter has filtered out all the images. How would you hope to fix the filter to give you what you want if you cannot see it at all? How would you know if you are still in the filtering mode if the text layer doesn't go back to viewer/gallery mode text upon ending the filtering?

This app is so much more than you are trying to convince everyone that it is. I understand that you may not want to add a billion of rarely used (but incredibly useful) features, but preventing them from being possible by not allowing the simplest of changes just seems odd. If it doesn't cost you anything, why prevent anyone from having it?

artemsen:
> That perhaps the user has

Don't base it on "perhaps". Until there are real use cases, I don't see the point of this overhaul. 

> Do we? If you were fine with a broken behaviour before, why would this differently-broken behaviour be an issue?

Previously (and in the current implementation), this field was not displayed if there was no image.

> If you were to implement an image filter

I’ve already said why I don’t want to do this.
If you use any undocumented methods to implement this functionality, bear in mind that it is a "dirty hack."
Now you are trying to bring these hacks into the master branch. Sorry, but no.

> If it doesn't cost you anything, why prevent anyone from having it?

If I had accepted every single proposal and whatever implementation idea happened to cross someone's mind, this app wouldn't be what it is today.

litoj:
> Don't base it on "perhaps". Until there are real use cases, I don't see the point of this overhaul.

What is so unreal about an image filter? I'm not the only one who wants it and it is a great feature that no other image viewer provides.

> Previously (and in the current implementation), this field was not displayed if there was no image.

How do I make a field _not_ display? When I clear the fields, I'd expect the line with the field to not display, but it does. What am I doing wrong then?

> If you use any undocumented methods...

How would I even do that? The api documents what the api exposes. I do wish that some things like filesystem events and imagelist changes were something I could listen to, but I can't.

>  bear in mind that it is a "dirty hack."

Yes, but that is not even possible, so who cares. What is a dirty hack that I had to employ is getting the cmdline of itself via reading `/proc`. What is a dirty hack is needing to defer a function whenever the user executes a command that contains `rm` or `mv` because I cannot listen to the fsevents that the internal eventloop already has anyway.

To be clear, I am not saying these are easy problems to solve (well... the first one is). I am just saying that talking about dirty hacks is unfair when you are the one requiring us to employ them.

All dirty hacks that I am doing I am doing because you counteract any 1st party solution without giving any constructive feedback, often stopping things with "because nobody wants this and you need a _real_ usecase", while the real usecase that started it you disregard anyway.

I get it, they are something that rarely anyone will use. But if it is a simple change vs a laborious and flaky workaround... Why are you making an amazing app at all then, when using it to its full potential is considered unwanted?

---

Can we be productive? What do you actually want me to change/do in order to fix yet another edge-case behaviour that any other programmer except you would regard as a bug? Text layer doesn't need any images to be fully working. The fields are nice but they can be almost as easily replaced with pure lua `viewer.get_image()[field]...`

The image isn't needed at all for the text layer to show text, so why do you want to prevent any text from being displayed when the imagelist is empty? Saying that the app has practically no purpose when the imagelist is empty is, first of all, not true at all (I can have a keybind for adding a directory via dmenu, etc.) and also it would be a reason to not implement the no-image display at all, as it was originally - but not to implement it half way!

IMHO either acknowledge a feature or don't allow it at all. Saying that there is a feature you refuse to fully implement/support and refuse to move forward any attempts at implementing the rest by third parties just feels like taunting the users who actually want to make the app complete and consistent. (not wanting new features is understandable. This isn't that though)

---

To be clear, I have noticed that displaying text on an empty imagelist is more difficult than I thought because you put so many roadblocks in place. Back when I suggested just using a dummy image for an empty imagelist you said stubs are not proper implementation. Well, may I just say, your implementation isn't proper either :/

I want to fix it and I will, but I don't see a reason why it would always have to be acompanied by arguing about why a half-assed feature should be an actual feature. You've already accepted it, so just let it be done all the way.

artemsen:
> The image isn't needed at all for the text layer to show text, so why do you want to prevent any text from being displayed when the imagelist is empty?

The text layer consists of information about the current image and the status string.
Changing this logic would require handling and hiding empty or invalid values ​​(as shown in the screenshots), resulting in work the purpose of which is unclear to me. As I understand it, you want to use text fields for keyboard input. To me, this looks like a dirty hack.

As long as I am the maintainer of this software, I take responsibility for its operation and reserve the right to accept or reject any changes.
If that doesn't suit you, feel free to develop your own fork — it's open source.

litoj:
> As I understand it, you want to use text fields for keyboard input.

No, I want it to use it for anything, not just for image displaying. That's all

It can list the current directory, marked images, filtered images, or some other computed value.

My whole point is that something generic shouldn't be arbitrarily bound to one specific thing and cripple the whole stack because of it.

And rather than apply a fix using 3 extra lines you will enforce yet another dirty hack on everyone that wants to use your app even when the imagelist is empty.

litoj:
> Changing this logic would require handling and hiding empty or invalid values

Why? It works just as well without it.

Your whole argument is that this would bring a lot of work that is not necessary. Whereas I am saying that why bother dealing with default values when there aren't any expected. The whole point of this is to make the text layer work at all, not to define default values for something that is literally undefined.

I want to apply a simple and good fix to a problem. You always keep expanding the problem beyond the horizon, even though the initial problem that is much more prominent didn't bother you at all to begin with. 

Just suddenly whenever a small problem caused by improper previous implementation comes up, you balloon it up to usecases which noone had nor cared about. Only you suddenly care and say that noone has such usecase to warrant such work, even though noone was asking for large changes.

litoj:
> As long as I am the maintainer of this software, I take responsibility for its operation and reserve the right to accept or reject any changes.
If that doesn't suit you, feel free to develop your own fork — it's open source.

I respect your rights fully. And your efforst as well. I just cannot respect your arguments, because they all seem very hypocritical - any change by you is fine even if it goes just half the way, as long as it shuts up an issue, but outsiders wanting to go one step further and get it to 2/3 now suddenly have to go all the way, even though _that_ actually wasn't needed for anyone, but the 2/3 were.

I just want to understand your decision logic so that I can align with it and perhaps do my implementations differently or something.

PR #577, opened 2026-09-10
ci: parallelized clang-tidy checks

litoj:
makes the static analysis faster by 2 minutes

### DIFF
--- a/.github/workflows/static_analisys.yml
+++ b/.github/workflows/static_analisys.yml
@@ -35,6 +35,24 @@ jobs:
         meson compile -C ${{ env.BUILD_DIR }}
 
     - name: Run analisys
-      run: >
-        docker exec swayimg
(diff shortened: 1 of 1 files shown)


artemsen:
This solution works only in ideal world.

1. Now CI has mandatory dependency to bash;
2. Now CI uses only system's clang-tidy instead of specified in meson;
3. Now CI uses hardcoded directory (that can be write-protected) to create temp files;
4. Now CI checks files that cannot be checked and skips files that can be checked.

litoj:
Ok, fair, this isn't a particularly fancy solution. And not a major speedup either. But I still need to know: What do you mean by

> Now CI checks files that cannot be checked and skips files that can be checked.

?

artemsen:
This script uses `*.cpp`, but it skip `test` dir; it checks `format/jpeg.cpp` even if it was removed from the build with `-D jpeg=disabled`.

PR #582, opened 2026-09-18
Add support for Krita's native file format.

talanis85:
Add a loader for Krita files.

[Krita](https://www.krita.org) is a popular open source digital painting program. Its native file format is basically a zipfile, that contains a file called `mergedimage.png` that can be used to preview the file contents.

See also: [https://docs.krita.org/en/general_concepts/file_formats/file_kra.html](https://docs.krita.org/en/general_concepts/file_formats/file_kra.html)

This loader uses the existing png loader to load the `mergedimage.png` from the zipfile. Zipfile loading is done by [minizip-ng](https://github.com/zlib-ng/minizip-ng), which is added as a dependency.

I'm not exactly fluent with C++ and meson so please excuse any mistakes or anti-patterns :)

I also tested with valgrind and found no leaks.

EDIT: Still trying to figure out github workflows. Hang on please.


artemsen:
Actually, Krita files can be opened in Swayimg without modifying the source code at all:
```
$ unzip -p image.kra mergedimage.png | swayimg -
```
So, I don't see much point in adding a new dependency just for unpacking an archive with png, but I'm not opposed to it.

You need to ensure that all tests pass: [migrate from C to C++](https://github.com/artemsen/swayimg/actions/runs/35399825238/job/105776927472?pr=582#step:7:90), [(may be) replace minizip with libz](https://github.com/artemsen/swayimg/actions/runs/35399825267/job/105776926930?pr=582#step:6:43), [fix memory leaks](https://github.com/artemsen/swayimg/actions/runs/35399825278/job/105776926726?pr=582#step:8:375) and so on. You can enable Actions support in your fork and use it to verify that the tests pass.

talanis85:
Yeah, the reason is that I would like to have these images in gallery mode too.

I used minizip-ng because it's more flexible when loading zip files from memory. The library seems to be pretty stable and part of all major distributions.

Tests now pass.

artemsen:
> I used minizip-ng because it's more flexible when loading zip files from memory. The library seems to be pretty stable and part of all major distributions.

I would prefer using zlib. It is included in all distributions and is almost always installed, unlike minizip-ng.

src/formats/krita.cpp
```diff
@@ -0,0 +1,119 @@
...
+    [[nodiscard]] ImagePtr decode(const Data& data) const override
+    {
+        void* mem_stream = nullptr;
+        void* zip_handle = nullptr;
+        int32_t err = MZ_OK;
+
+        mem_stream = mz_stream_mem_create();
+        mz_stream_mem_set_buffer(mem_stream, data.data, data.size);
+
+        if (MZ_OK != mz_stream_open(mem_stream, nullptr, MZ_OPEN_MODE_READ)) {
```
artemsen, line 28: Please don't use yoda-style.

src/formats/krita.cpp
```diff
@@ -0,0 +1,119 @@
...
+            mz_stream_mem_delete(&mem_stream);
+            return nullptr;
+        }
+
+        zip_handle = mz_zip_create();
+
+        if (MZ_OK != mz_zip_open(zip_handle, mem_stream, MZ_OPEN_MODE_READ)) {
+            mz_zip_delete(&zip_handle);
```
artemsen, line 36: Use unque_ptr for such objects. [Here](https://github.com/artemsen/swayimg/blob/60158bfa75e45ace09fb8d5d75c1d6914a72e750/src/formats/tiff.cpp#L78) is example.

src/formats/krita.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "mz_strm.h"
+#include "mz_strm_mem.h"
+#include "mz_zip.h"
+
+namespace {
+
+class ImageFormatKrita : public ImageFormat {
+public:
+    ImageFormatKrita() noexcept
+        : ImageFormat(Priority::Normal, "kra")
```
artemsen, line 15: I think it's very rare format, so it can has low priority.

src/formats/krita.cpp
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    [[nodiscard]] ImagePtr decode(const Data& data) const override
+    {
+        void* mem_stream = nullptr;
+        void* zip_handle = nullptr;
+        int32_t err = MZ_OK;
+
+        mem_stream = mz_stream_mem_create();
+        mz_stream_mem_set_buffer(mem_stream, data.data, data.size);
```
artemsen, line 26: You should check that the stream is really in ZIP format before opening it.

src/formats/krita.cpp
```diff
@@ -0,0 +1,119 @@
...
+            mz_stream_mem_delete(&mem_stream);
+            return nullptr;
+        }
+
+        int32_t bytes_read;
+        do {
+            uint8_t* newdata = nullptr;
+
+            newdata = reinterpret_cast<uint8_t*>(
+                realloc(png_data.data, png_data.size + 4096));
```
artemsen, line 72: Use std::vector instead.

talanis85:
Thanks for the review!

I will rewrite it with zlib, hang on please. For now, I pushed some changes where I remove the Yoda-talk, use vector for the reading of the data and unique_ptr for zip_handle and mem_stream. Maybe you could have a quick look if I used these correctly. Thank you!

artemsen:
The archive contains an embedded preview.png; I think we can use it, as it will speed up thumbnail loading in the gallery.
[Here](https://github.com/artemsen/swayimg/blob/f19e2cf0c4c37b74204a1e4adc41a2c3fe65dfb1/src/formats/raw.cpp#L78) is an example as it done for RAW images with embedded thumbnail.

